PR #17, opened 2021-01-19
Make webpack script more readable

BakiVernes:
Made the install script shorter and easier to read with a stylesheets template


dhh:
Included this change in a bigger clean up just committed 👍

Issue #20, opened 2021-01-19
group-hover classes are not removed


andersklenke:
After https://github.com/rails/tailwindcss-rails/commit/8af7bfcb1a11d7cd63a3be4a6a3c53a9a37df679 group-hover classes, `group-hover:text-gray-900` for example, are not purged. 

This adds a substantial amount of classes to the outputted css. 

Unfortunately my regex skills are not strong enough to fix it.

(Originally created a [comment](https://github.com/rails/tailwindcss-rails/issues/15) but a new issue is probably more visible to people. )

kaspth:
Possible to open a PR with a failing test? Can try to fiddle with rubular.com to help sift the regex tweaks needed.

andersklenke:
Sure! Will open a failing test pr tomorrow

PR #22, opened 2021-01-20
Add test for group-hover classes

andersklenke:
This adds a failing test for group hover classes as discussed in here https://github.com/rails/tailwindcss-rails/issues/20

cc. @kaspth

### DIFF
--- a/test/fixtures/group_hover.html.erb
+++ b/test/fixtures/group_hover.html.erb
@@ -1,0 +1,4 @@
+<div class="group border-indigo-500 hover:bg-white hover:shadow-lg hover:border-transparent">
+  <p class="text-indigo-600 group-hover:text-gray-900">New Project</p>
+  <p class="text-indigo-500 group-hover:text-gray-500">Create a new project from a variety of starting templates.</p>
+</div>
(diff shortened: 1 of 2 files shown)


kaspth:
So it feels like the focus/hover purging is normalizing the passed in CSS class such that `.group-hover` looks like it's being used even though it isn't https://github.com/rails/tailwindcss-rails/blob/1ff4d8fd817f81228cfdb8a39643ee92e6ef1bd4/lib/tailwindcss/purger.rb#L57

andersklenke:
I believe this happens due to the `.group` class in the parent class. The word `group` will match with every single class that contains "group" afterwards. 
https://github.com/rails/tailwindcss-rails/blob/63c6058840032387878f1bd04e1f0c957364e446/test/fixtures/group_hover.html.erb#L1

kaspth:
Ah! So we may need to throw a \b or \s in the final regex, to check for a word boundary or whitespace, respectively. Wanna try fiddling with either of those?

andersklenke:
Sure! I will give it a try

andersklenke:
A lot happened for me lately, so I haven't had the time to look at this yet. If anyone want to pick it up, feel free 🙂

PR #33, opened 2021-02-04
Webpacker: Add tailwind.config.js

dixpac:
During webpack installation generate `tailwind.config.js`.

This commit places that file inside `stylesheets/` dir, and link to it
inside `postcss.config.js`.

This is the blank Tailwind confg file that one can extend
to bring its own esthetics.


ghiculescu:
What do you think about enabling purging in this file? That way the webpacker behavior would match the asset pipeline [behavior](https://github.com/rails/tailwindcss-rails#purging-in-production) out of the box.

I think the change would just be this:

```diff
-  purge: [],
+  purge: {
+    enabled: process.env.NODE_ENV === "production"
+    content: ['./app/views/**/*.html.erb'],
+  },
```

But wouldn't mind someone sanity checking that.

dixpac:
@ghiculescu something like this https://github.com/rails/tailwindcss-rails/pull/40 ?

ghiculescu:
@dixpac yes, that's great!

Issue #31, opened 2021-02-03
Can the class name pattern regex be adjusted to allow for HAML templates?


wnm:
Hi there 👋

I was wondering what the reason is for having a dot in the regex to search for class names? Are there tailwind class names that include a dot?

The reason I'm asking is, because this way HAML templates don't work:
https://regex101.com/r/hiAKC7/2

Without dots, it finds all class names for both html and haml templates:
https://regex101.com/r/0JJDKQ/1

Except I'm missing something?

dhh:
If you can do a PR with enough tests to verify that everything else is still working, then have at it!

jumph4x:
Ooooh, this is cool. I'll try to take a stab at it in a little bit.

wnm:
I've opened a PR here: #38

PR #38, opened 2021-02-07
Change CLASS_NAME_PATTERN to work also for haml templates

wnm:
This PR changes the CLASS_NAME_PATTERN to work for haml templates.

Since haml syntax uses dots to chain class names, I first removed the dot from the search pattern...

But, to make sure it still works for decimal spacing classes (e.g. `my-1.5`), I added a pattern to allow dots which are preceded by a number between 0-3 and followed by a 5.

By looking through the tailwind documentation, I think all the different class patterns are covered by a test already, so I'm not sure how to improve the test coverage... I'm happy to add more tests, but would need some input first, on what to test for...

### DIFF
--- a/lib/tailwindcss/purger.rb
+++ b/lib/tailwindcss/purger.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class Tailwindcss::Purger
-  CLASS_NAME_PATTERN = /[:A-Za-z0-9_-]+[\.]*[\\\/:A-Za-z0-9_-]*/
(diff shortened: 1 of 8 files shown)


jumph4x:
Let's link this to #31, I almost got started on this not realizing you made moves already. 

Also, thanks for tackling it. <3

wnm:
ugh, sorry about that I totally forgot to mention this PR in a comment in #31. I don't think I have the permission to link the issue as well...

jontebol:
@wnm, do you think this will also work for [slim templates](http://slim-lang.com/)?

wnm:
yes, it should work, at least for the default configuration which is using the same syntax for class names as haml, e.g: `.class-name`

dhh:
Happy to see this adapted to the new purger.

wnm:
@dhh done! I also added tests for slim syntax. it has a similar syntax (the same regex works), but [better tailwind support](https://github.com/slim-template/slim/pull/841/files)

krskibin (commented):
:+1:

PR #39, opened 2021-02-07
Faster webpack compile time

codingluke:
Hi there,

when adding custom CSS into a separate file, webpack should not have to compile the whole tailwindcss again, but only the changes in `application.scss`.

Moreover, I personally like it to have the custom CSS in a separate file and not in between the tailwind includes.

See: https://github.com/tailwindlabs/tailwindcss/issues/443#issuecomment-735583969

**Before**
grafik

**After**
grafik

I hope this will help others too. 😄

### DIFF
--- a/lib/install/stylesheets/components.scss
+++ b/lib/install/stylesheets/components.scss
@@ -1,0 +1,4 @@
+@import "tailwindcss/components";
+
+// Import components added to `./components/` here
+// https://tailwindcss.com/docs/extracting-components
--- a/lib/install/stylesheets/application.scss
+++ b/lib/install/stylesheets/application.scss
@@ -1,3 +1,0 @@
-@import "tailwindcss/base";
-@import "tailwindcss/components";
-@import "tailwindcss/utilities";
--- a/lib/install/stylesheets/base.scss
+++ b/lib/install/stylesheets/base.scss
@@ -1,0 +1,4 @@
+@import "tailwindcss/base";
+
+// Add custom base styles here
+// https://tailwindcss.com/docs/adding-base-styles
--- a/lib/install/stylesheets/utilities.scss
+++ b/lib/install/stylesheets/utilities.scss
@@ -1,0 +1,4 @@
+@import "tailwindcss/utilities";
+
+// Import vendor CSS here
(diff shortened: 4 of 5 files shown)


dixpac:
This is awesome @codingluke! Indeed on our app this speeds up compilation time ~3x.

Maybe we could organize the files a bit differently such as:

```
stylesheets/
  components/
  base.scss
  components.scss
  tailwind.config.js
  utilities.css
```

```scss
// base.scss

// This injects Tailwind's base styles, which is a combination of Normalize.css and additional base styles.
@import 'tailwindcss/base';

// Add custom base styles, applied after the tailwind-base classes
```

```scss
// components.scss
`@import 'tailwindcss/components';`
  
 // Custom components....
```

```scss
// utilities.scss

// This injects all of Tailwind's utility classes, generated based on your config file.
@import 'tailwindcss/utilities';

// Include vendor css.
```


```js
// application.js

// These are imported separately for faster Webpack recompilation
import "stylesheets/base.scss"
import "stylesheets/components.scss"
import "stylesheets/utilities.scss"

```

codingluke:
Hi @dixpac, Thanks for your input :) 

I see your the point! However, isn't `application.scss` kind of a convention name in Rails? Also I am not sure if the custom CSS can go directly after the `@import "tailwindcss/base". I have to check the benchmarks first. 
Anyways, before I spend too much time, I would like to have @dhh feedback for the naming as naming is hard ;)

codingluke:
@dixpac, I just checked the benchmarks, it is totally fine with your structure too 👍  So when everybody can confirm on this I will adapt it.

lib/install/stylesheets/components.scss
```diff
@@ -0,0 +1,4 @@
+@import "tailwindcss/components";
+
+// Import components added to `./components/` here
+// https://tailwindcss.com/docs/extracting-components
```
codingluke, line 4: @dixpac why do we need the special components folder? There is also no folder to structure the base styles (and I would do this too).

I see two ways here:

1. Everybody should add the own structure
2. We give them a structure (CoC). Just to do that, I am not too much experienced with tailwind ;)

PR #40, opened 2021-02-08
Webpacker: Basic purging patterns

dixpac:
This PR configures Tailwind purging with basic rails app needs.
Developers can easily change this down the road, but this should serve
as a good starting point.

-----

On bare-bones rails new (_running this in development env_):

**Before**
Screenshot 2021-02-08 at 14 48 41

**After**
Screenshot 2021-02-08 at 14 49 36

ps
In real-world app difference will be smaller, since there is less to purge :)

### DIFF
--- a/lib/install/stylesheets/tailwind.config.js
+++ b/lib/install/stylesheets/tailwind.config.js
@@ -1,0 +1,19 @@
+module.exports = {
+  // Purge unused TailwindCSS styles
+  purge: {
+    enabled: ["production"].includes(process.env.NODE_ENV),
+    content: [
+      './**/*.html.erb',
+      './app/helpers/**/*.rb',
+      './app/javascript/**/*.js',
+    ],
+  },
+  darkMode: false, // or 'media' or 'class'
+  theme: {
+    extend: {},
+  },
+  variants: {
+    extend: {},
+  },
+  plugins: [],
+}
--- a/lib/install/tailwindcss_with_webpacker.rb
+++ b/lib/install/tailwindcss_with_webpacker.rb
@@ -7,7 +7,6 @@ insert_into_file "#{Webpacker.config.source_entry_path}/application.js", "\nimpo
 
 say "Configuring Tailwind CSS"
 directory Pathname.new(__dir__).join("stylesheets"), Webpacker.config.source_path.join("stylesheets")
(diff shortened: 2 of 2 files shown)


lib/install/tailwindcss_with_webpacker.rb
```diff
@@ -7,7 +7,6 @@
 
 say "Configuring Tailwind CSS"
 directory Pathname.new(__dir__).join("stylesheets"), Webpacker.config.source_path.join("stylesheets")
-Dir.chdir(WEBPACK_STYLESHEETS_PATH) { run "npx tailwindcss init" }
```
codingluke, line 10: Is this not needed anymore? Or what is the connection from this line to the purging config of this PR?

dixpac: Yup not needed anymore, since I moved `tailwind.config.js` to "templates", reason is purge configuration.
It's being copied on the line above(9) together with `application.css`.

dixpac:
@dhh wdyt?

lib/install/stylesheets/tailwind.config.js
```diff
@@ -0,0 +1,19 @@
...
+  // Purge unused TailwindCSS styles
+  purge: {
+    enabled: ["production"].includes(process.env.NODE_ENV),
+    content: [
+      './**/*.html.erb',
+      './app/helpers/**/*.rb',
+      './app/javascript/**/*.js',
+    ],
+  },
+  darkMode: false, // or 'media' or 'class'
```
ghiculescu, line 11: unrelated to this PR, but should this be updated to match https://github.com/rails/tailwindcss-rails/blob/main/tailwind.config.js#L4 ?

lib/install/stylesheets/tailwind.config.js
```diff
@@ -0,0 +1,19 @@
+module.exports = {
+  // Purge unused TailwindCSS styles
+  purge: {
+    enabled: ["production"].includes(process.env.NODE_ENV),
+    content: [
+      './**/*.html.erb',
+      './app/helpers/**/*.rb',
```
ghiculescu, line 7:
```suggestion
      './app/views/**/*.*,
      './app/helpers/**/*.rb',
```

IMO this should match the compressor: https://github.com/rails/tailwindcss-rails/blob/main/lib/tailwindcss/compressor.rb#L14, so that either way you use this gem you get the same outcome.

PR #41, opened 2021-02-09
Use path from webpacker config instead hardcoded path

Zhong-z:
I couldn't find a public method for getting [Webpacker config](https://github.com/rails/webpacker/blob/master/lib/webpacker/configuration.rb) relative path, so I just substitute Rails root path with empty string.


dhh:
This gem no longer supports a webpack installation, so this isn't relevant any longer. Will work on a different way to make it easy to add TW to a webpack setup.

PR #63, opened 2021-08-16
Skip sprockets initializers for apps that don't use sprockets

gmcgibbon:
Some apps don't use sprockets and the current set of initializers assumes sprockets to be loaded. I think it would be reasonable to conditionally set initializers based on if `Sprockets` is defined or not. This should be done before the gem is required in sprockets railtie that rails loads with "rails/all":

https://github.com/rails/rails/blob/13a714f30d5e4c4a53062bb1b29fbeb287efd3be/railties/lib/rails/all.rb#L18
https://github.com/rails/sprockets-rails/blob/118ce60b1ffeb7a85640661b014cd2ee3c4e3e56/lib/sprockets/railtie.rb#L6

Also, the first commit is to clean up the mimemagic reference in the lockfile.


dhh:
This gem has now doubled down on being a solutions specifically for sprockets based apps, so I don't think this is relevant any longer.

PR #64, opened 2021-08-17
Remove webpacker support

dixpac:
The heart of this engine is a sprockets-specific approach to using TW with the asset pipeline.
By removing the webpacker support we can focus our efforts in building a
better TW asset pipeline approach.

Currently, if you need advanced TW configuration you will need to go
through the webpack(or other js bundlers).

r? @dhh


dixpac:
Fixed mimemagic reference the make a CI ✅

PR #67, opened 2021-08-18
Make the engine Rails 5+ compatible

dixpac:
This PR makes the engine compatible with rails 5+ versions.


dixpac:
Oh the CI 😢 , not sure if this was the smartest idea 😄

PR #70, opened 2021-08-31
Commit auto-updates only if assets have changed

jonathanhefner:
Follow-up to #48.

This prevents commits that only update `yarn.lock`, such as 2338bd6187da1f588903b2ba2f1d2692a654d6a3.

### DIFF
--- a/.github/workflows/update-tailwind.yml
+++ b/.github/workflows/update-tailwind.yml
@@ -21,7 +21,7 @@ jobs:
       run: |
         git config --local user.email "github-actions@example.com"
         git config --local user.name "GitHub Actions"
-        git commit -am "Update tailwind.css" || echo "No changes to commit"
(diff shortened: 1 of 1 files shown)


.github/workflows/update-tailwind.yml
```diff
@@ -21,7 +21,7 @@ jobs:
       run: |
         git config --local user.email "github-actions@example.com"
         git config --local user.name "GitHub Actions"
-        git commit -am "Update tailwind.css" || echo "No changes to commit"
+        git diff --quiet app/assets || git commit -am "Update tailwind.css"
```
jonathanhefner, line 24: This uses `||` (not `&&`) because [`git diff --quiet`](https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---quiet) returns a truthy exit code when there are no differences.  (In shell scripting, an exit code of `0` is truthy and `1` is falsy.)

Issue #23, opened 2021-01-22
Ship scaffold generator that uses a nice Tailwind design


dhh:
The default Rails scaffold is plain to a fault. Let's ship some scaffold templates with this gem that uses Tailwind to demonstrate it.

gobijan:
Okay just to get the discussion started I think we first need to agree on a few things:

1. Do we want to use table or div under the hood?
**Divs** make it easier to adjust to mobile and also for turbo-rails to substitute responses if I remember correctly.
**Tables** are more structured and offer some kind of easy to add sortability (for example https://datatables.net or https://www.w3schools.com/w3js/w3js_sort.asp or whatever).

2. Do we want to show all properties by default or say only the first 4 (or whatever number)?
If we show all tables might really be the wrong approach given they are a pain on mobile.
Giving it a thought definition lists might be a feasible approach. Any opinion on that?

3. What about the actions per item?
Do we want to show all actions on every item or do we want to make an item clickable and show the edit or delete functionality only on the details page? This would provide less visual clutter.

4. Do we want to add javascript functionality out of the box?
Not sure if we want to use dropdowns with additional actions like edit or delete. I would say no but let me know what you think.

5. Do we have any color preferences? 
Rails uses a lot of red but for example a create item button is in western color usage more of a dangerous thing.

I played around a little and came up with this. It is still not nearly refined but uses definition list under the hood as an idea.
![Bildschirmfoto 2021-01-24 um 22 48 13](https://user-images.githubusercontent.com/49339/105644800-cbf78100-5e97-11eb-8f01-c794fc242290.png)

Tailwind UI gave me the idea of stacked lists BUT how to choose what to show or not, when there are many items?
See screenshot:
![Bildschirmfoto 2021-01-24 um 23 00 32](https://user-images.githubusercontent.com/49339/105644845-2bee2780-5e98-11eb-8225-00dab290c9e4.png)

hahmed:
@gobijan maybe worth creating a PR that adds the scaffolds and the basic styles then the discussion can happen on the PR?

We should be showing all the records by default and perhaps skip the colours in the initial PR (links commonly have blue so that should be good?)

dixpac:
@dhh you want this scaffold generator to override default rails generator(`bin/rails g scaffold`), or with different name such as `bin/rails g tailwind_scaffold ...`?

dhh:
I'd like to see it overridden. If you have tailwind installed, you want tailwind scaffolds.

[...]

PR #71, opened 2021-09-03
Add Tailwind styled scaffold templates

BakiVernes:
## Summary

This PR adds Tailwind styled scaffold templates when this gem is installed. These styles were sanctioned by Adam and the Tailwind team. Some examples:

Screenshot 2021-09-03 at 09 44 50
Screenshot 2021-09-03 at 09 44 03
Screenshot 2021-09-03 at 09 43 49

## Additional information

It would great if we had the option of using Tailwinds `@apply` directive. This would be useful when extracting components like buttons, inputs etc. But since the css comes precompiled we did what we could. 

Maybe something to work on in the future?

Co-authored-by: Dino Marić <dino.onex@gmail.com>

### DIFF
--- a/lib/install/tailwindcss.rb
+++ b/lib/install/tailwindcss.rb
@@ -1,11 +1,13 @@
 APPLICATION_LAYOUT_PATH = Rails.root.join("app/views/layouts/application.html.erb")
 
 if APPLICATION_LAYOUT_PATH.exist?
-  say "Add Tailwindcss include tags in application layout"
+  say "Add Tailwindcss include tags and container element in application layout"
   insert_into_file APPLICATION_LAYOUT_PATH.to_s, <<~ERB.indent(4), before: /^\s*<%= stylesheet_link_tag/
     <%= stylesheet_link_tag "inter-font", "data-turbo-track": "reload" %>
     <%= stylesheet_link_tag "tailwind", "data-turbo-track": "reload" %>
   ERB
+  insert_into_file APPLICATION_LAYOUT_PATH.to_s, %(    <main class="container mx-auto mt-28 px-5 flex">\n  ), before: /^\s*<%= yield/
+  insert_into_file APPLICATION_LAYOUT_PATH.to_s, %(\n    </main>),  after: /^\s*<%= yield %>/
 else
   say "Default application.html.erb is missing!", :red
   say %(        Add <%= stylesheet_link_tag "inter-font", "data-turbo-track": "reload" %> and <%= stylesheet_link_tag "tailwind", "data-turbo-track": "reload" %> within the <head> tag in your custom layout.)
--- a/lib/templates/erb/scaffold/_form.html.erb.tt
+++ b/lib/templates/erb/scaffold/_form.html.erb.tt
@@ -1,0 +1,43 @@
+<%%= form_with(model: <%= model_resource_name %>, class: "contents") do |form| %>
+  <%% if <%= singular_table_name %>.errors.any? %>
+    <div id="error_explanation" class="bg-red-50 text-red-500 px-3 py-2 font-medium rounded-lg mt-3">
+      <h2><%%= pluralize(<%= singular_table_name %>.errors.count, "error") %> prohibited this <%= singular_table_name %> from being saved:</h2>
+
+      <ul>
+        <%% <%= singular_table_name %>.errors.each do |error| %>
+          <li><%%= error.full_message %></li>
+        <%% end %>
+      </ul>
+    </div>
+  <%% end %>
+
+<% attributes.each do |attribute| -%>
+  <div class="field my-5">
+<% if attribute.password_digest? -%>
+    <%%= form.label :password %>
+    <%%= form.password_field :password %>
+</div>
+
+<div class="field my-5">
+    <%%= form.label :password_confirmation %>
+    <%%= form.password_field :password_confirmation, class: "block shadow rounded-md border border-gray-200 outline-none px-3 py-2 mt-2 w-full" %>
+<% elsif attribute.attachments? -%>
+    <%%= form.label :<%= attribute.column_name %> %>
+    <%%= form.<%= attribute.field_type %> :<%= attribute.column_name %>, multiple: true, class: "block shadow rounded-md border border-gray-200 outline-none px-3 py-2 mt-2 w-full" %>
+<% else -%>
+    <%%= form.label :<%= attribute.column_name %> %>
+<% if attribute.field_type == :text_area -%>
+    <%%= form.<%= attribute.field_type %> :<%= attribute.column_name %>, rows: 4, class: "block shadow rounded-md border border-gray-200 outline-none px-3 py-2 mt-2 w-full" %>
+<% elsif attribute.field_type == :check_box -%>
+    <%%= form.<%= attribute.field_type %> :<%= attribute.column_name %>, class: "block mt-2 h-5 w-5" %>
+<% else -%>
+    <%%= form.<%= attribute.field_type %> :<%= attribute.column_name %>, class: "block shadow rounded-md border border-gray-200 outline-none px-3 py-2 mt-2 w-full" %>
+<% end -%>
+<% end -%>
+  </div>
+
+<% end -%>
+  <div class="actions inline">
+    <%%= form.submit class: "rounded-lg py-3 px-5 bg-blue-600 text-white inline-block font-medium" %>
+  </div>
+<%% end %>
--- a/lib/templates/erb/scaffold/edit.html.erb.tt
+++ b/lib/templates/erb/scaffold/edit.html.erb.tt
@@ -1,0 +1,8 @@
+<div class="mx-auto md:w-2/3 w-full">
+  <h1 class="font-bold text-4xl">Editing <%= human_name.downcase %></h1>
+
+  <%%= render "form", <%= singular_table_name %>: @<%= singular_table_name %> %>
+
+  <%%= link_to "Show this <%= human_name.downcase %>", @<%= singular_table_name %>, class: "ml-2 rounded-lg py-3 px-5 bg-gray-100 inline-block font-medium" %>
+  <%%= link_to "Back to <%= human_name.pluralize.downcase %>", <%= index_helper %>_path, class: "ml-2 rounded-lg py-3 px-5 bg-gray-100 inline-block font-medium" %>
+</div>
--- a/lib/templates/erb/scaffold/new.html.erb.tt
+++ b/lib/templates/erb/scaffold/new.html.erb.tt
@@ -1,0 +1,7 @@
+<div class="mx-auto md:w-2/3 w-full">
+  <h1 class="text-lg font-bold text-4xl">New <%= human_name.downcase %></h1>
+
+  <%%= render "form", <%= singular_table_name %>: @<%= singular_table_name %> %>
+
+  <%%= link_to 'Back to <%= human_name.pluralize.downcase %>', <%= index_helper %>_path, class: "ml-2 rounded-lg py-3 px-5 bg-gray-100 inline-block font-medium" %>
+</div>
--- a/lib/tailwindcss/engine.rb
+++ b/lib/tailwindcss/engine.rb
@@ -19,5 +19,9 @@ module Tailwindcss
         env.cache = ActiveSupport::Cache.lookup_store(:null_store)
       end if Rails.env.production?
     end
+
+    config.app_generators do |g|
+      g.templates.unshift File::expand_path('../../templates', __FILE__)
+    end
   end
 end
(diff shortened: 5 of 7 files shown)


BakiVernes:
Resolves #23

lib/install/tailwindcss.rb
```diff
@@ -1,11 +1,13 @@
 APPLICATION_LAYOUT_PATH = Rails.root.join("app/views/layouts/application.html.erb")
 
 if APPLICATION_LAYOUT_PATH.exist?
-  say "Add Tailwindcss include tags in application layout"
+  say "Add Tailwindcss include tags and container element in application layout"
   insert_into_file APPLICATION_LAYOUT_PATH.to_s, <<~ERB.indent(4), before: /^\s*<%= stylesheet_link_tag/
     <%= stylesheet_link_tag "inter-font", "data-turbo-track": "reload" %>
     <%= stylesheet_link_tag "tailwind", "data-turbo-track": "reload" %>
   ERB
+  insert_into_file APPLICATION_LAYOUT_PATH.to_s, %(    <main class="container mx-auto mt-28 px-5 flex">\n  ), before: /^\s*<%= yield/
```
dixpac, line 9: @dhh take a look on this also when you have some time https://github.com/rails/rails/pull/43067 🙏🏼

dhh:
Digging this! Will look closer in a bit. Thanks for working on it!

lib/templates/erb/scaffold/_form.html.erb.tt
```diff
@@ -0,0 +1,43 @@
...
+      <ul>
+        <%% <%= singular_table_name %>.errors.each do |error| %>
+          <li><%%= error.full_message %></li>
+        <%% end %>
+      </ul>
+    </div>
+  <%% end %>
+
+<% attributes.each do |attribute| -%>
+  <div class="field my-5">
```
hahmed, line 15: `field` what is the field attribute? Is this the scaffold field class here https://github.com/rails/rails/blob/9c77d72ecc7fdf490e778b0807882fd53115e7a3/actiontext/test/dummy/app/assets/stylesheets/scaffold.css#L40? If so, we can drop that because we want to use only tailwind inlined css?

lib/templates/erb/scaffold/_form.html.erb.tt
```diff
@@ -0,0 +1,43 @@
...
+<% elsif attribute.field_type == :check_box -%>
+    <%%= form.<%= attribute.field_type %> :<%= attribute.column_name %>, class: "block mt-2 h-5 w-5" %>
+<% else -%>
+    <%%= form.<%= attribute.field_type %> :<%= attribute.column_name %>, class: "block shadow rounded-md border border-gray-200 outline-none px-3 py-2 mt-2 w-full" %>
+<% end -%>
+<% end -%>
+  </div>
+
+<% end -%>
+  <div class="actions inline">
```
hahmed, line 40: `actions` also looks like it's from the default scaffold css from https://github.com/rails/rails/blob/9c77d72ecc7fdf490e778b0807882fd53115e7a3/actiontext/test/dummy/app/assets/stylesheets/scaffold.css#L41, maybe drop this too and add mb-something here too?

lib/templates/erb/scaffold/_form.html.erb.tt
dixpac: Ooops...fixed

lib/templates/erb/scaffold/_form.html.erb.tt
dixpac: Ooops...fixed. Thank you ❤️

dhh:
We have to make this match the new style in rails/main where the _post.html.erb partial is used on the index. That's how it matches up to work out of the box with Turbo. Which means we can't use a table on the index.

See the templates in https://github.com/rails/rails/tree/main/railties/lib/rails/generators/erb/scaffold/templates

dixpac:
@dhh no problem.
That was the one of the points we debated internally, main reason why we choose table is that Adam sent us mockups with the table and we didn't wan't to deviate too much from his mockups.

But, yeah I agree that index without the table is going to be easier especially for the scaffolds where the table columns are dynamic and it is hard to style the table properly. 

Anyways, we'll match with the Rails 7 changes 👍🏻

lib/templates/erb/scaffold/_form.html.erb.tt
```diff
@@ -0,0 +1,43 @@
...
+  <%% end %>
+
+<% attributes.each do |attribute| -%>
+  <div class="my-5">
+<% if attribute.password_digest? -%>
+    <%%= form.label :password %>
+    <%%= form.password_field :password %>
+</div>
+
+<div class="my-5">
```
dhh, line 21: Don't think these div's are aligned properly, yeah?

lib/templates/erb/scaffold/edit.html.erb.tt
```diff
@@ -0,0 +1,8 @@
+<div class="mx-auto md:w-2/3 w-full">
+  <h1 class="font-bold text-4xl">Editing <%= human_name.downcase %></h1>
+
+  <%%= render "form", <%= singular_table_name %>: @<%= singular_table_name %> %>
+
+  <%%= link_to "Show this <%= human_name.downcase %>", @<%= singular_table_name %>, class: "ml-2 rounded-lg py-3 px-5 bg-gray-100 inline-block font-medium" %>
+  <%%= link_to "Back to <%= human_name.pluralize.downcase %>", <%= index_helper %>_path, class: "ml-2 rounded-lg py-3 px-5 bg-gray-100 inline-block font-medium" %>
+</div>
```
dhh, line 8: Add CR

lib/templates/erb/scaffold/new.html.erb.tt
```diff
@@ -0,0 +1,7 @@
+<div class="mx-auto md:w-2/3 w-full">
+  <h1 class="text-lg font-bold text-4xl">New <%= human_name.downcase %></h1>
+
+  <%%= render "form", <%= singular_table_name %>: @<%= singular_table_name %> %>
+
+  <%%= link_to 'Back to <%= human_name.pluralize.downcase %>', <%= index_helper %>_path, class: "ml-2 rounded-lg py-3 px-5 bg-gray-100 inline-block font-medium" %>
+</div>
```
dhh, line 7: Add CR

lib/templates/erb/scaffold/partial.html.erb.tt
```diff
@@ -0,0 +1,22 @@
...
+<% end -%>
+  </p>
+
+<% end -%>
+  <%% if action_name != "show" %>
+    <%%= link_to "Show this <%= human_name.downcase %>", <%= singular_table_name %>, class: "rounded-lg py-3 px-5 bg-gray-100 inline-block font-medium" %>
+    <%%= link_to 'Edit this <%= human_name.downcase %>', edit_<%= singular_table_name %>_path(<%= singular_table_name %>), class: "rounded-lg py-3 ml-2 px-5 bg-gray-100 inline-block font-medium" %>
+    <hr class="mt-6">
+  <%% end %>
+</div>
```
dhh, line 22: Add CR

lib/templates/erb/scaffold/_form.html.erb.tt
BakiVernes: When we generate the scaffold the divs are indeed properly aligned as shown in a fresh rails/main app:

image

lib/generators/tailwindcss/scaffold/scaffold_generator.rb
```diff
@@ -0,0 +1,34 @@
...
+
+      def copy_view_files
+        available_views.each do |view|
+          formats.each do |format|
+            filename = filename_with_extensions(view, format)
+            template filename, File.join("app/views", controller_file_path, filename)
+          end
+        end
+
+        template "partial.html.erb", File.join("app/views", controller_file_path, "_#{singular_table_name}.html.erb")
```
dixpac, line 25: In order to get the partial in rails 6 I had to "re-implement" Rails 7 scaffold generator :)

dixpac:
@dhh I think this should be ready now, and it should also work with Rails 6. Take a 👀 when you have some time.

dhh:
Finally got a chance to check this out. Looking very good! Only thing I'd say is that we could use some padding. So headings and buttons aren't slammed against the edges of the browser:

Screen Shot 2021-09-28 at 8 30 35 AM

lib/generators/tailwindcss/scaffold/scaffold_generator.rb
```diff
@@ -0,0 +1,34 @@
...
+      end
+
+      private
+        def available_views
+          %w(index edit show new _form)
+        end
+    end
+  end
+end
```
dhh, line 34: Add CR

test/lib/generators/tailwindcss/scaffold_generator_test.rb
```diff
@@ -0,0 +1,21 @@
...
+  Minitest.after_run do
+    FileUtils.rm_rf GENERATION_PATH
+  end
+
+  test "generates correct view templates" do
+    run_generator
+
+    %w(index edit new show _form _message).each { |view| assert_file "app/views/messages/#{view}.html.erb" }
+  end
+end
```
dhh, line 21: Add CR

dixpac:
@dhh re paddings that is strange! Only way I could see that happening is if you don't have a `main` wrapper(https://github.com/rails/tailwindcss-rails/pull/71/files#diff-82d31adf98ad40bfa67ed05464d36790777703adead66a2478d97320061f136cR9)

Could you push sample app you are showing on the Github so we can double check 🙏

For the reference here how it looks like on the app I've just created

Screenshot 2021-09-28 at 20 18 39

dhh:
I had not run the installer correctly. You're right. It looks great!! This is really neat.

lib/templates/erb/scaffold/_form.html.erb.tt
Willardgmoore: I have had to add that extra indentation for  `.tt` files that I have worked on in the past too so that the generated files look right as you showed above. That's not the only quirk it has for lining up code. Variables and things also have to be lined up weird so they generate in the correct location. I have never understood why, but would like to look into it more. Any suggestions on where I could dive deeper on this?

seanwmitchell:
@dixpac Thanks for your work on this, I’m finally going to get rid of rails_admin gem (with heaps of dependent gems) and scaffold out the rest of my models. Great work 👍 I appreciate it.

lafeber:
@BakiVernes I absolutely love this, I was working on a similar thing :) Have you considered applying the tailwind classes in css, like daisy-ui does?

BakiVernes:
Thank you @lafeber! We have considered extracting classes, but since this gem comes with a precompiled css file with all of tailwinds utility classes inside (so no preprocessing done) it's gonna be a challenge. But we will definitely take a shot at it :)

lafeber:
I'm still not sure what the best DRY approach is for Rails. Viewcomponents in combination with helper methods like `def input_class` ?

esale:
> We have to make this match the new style in rails/main where the _post.html.erb partial is used on the index. That's how it matches up to work out of the box with Turbo. Which means we can't use a table on the index.
> 
> See the templates in https://github.com/rails/rails/tree/main/railties/lib/rails/generators/erb/scaffold/templates

I'm sorry if this is not the place for this comment:

@dhh I love Rails, among other things, because it enables me to do very fast functional prototypes. And part of that was having an index scaffold that I could instantly show to a client to display a list of items, and customize later. It was amazing and well known:
**Rails automates scaffolding to make early coding more productive than ever before.** ([Ruby on Rails: Up and Running by Bruce Tate, Curt Hibbs](https://learning.oreilly.com/library/view/ruby-on-rails/0596101325/))

Now, with Rails 7.0.3, if I wanted that, do I have to manually code another view? Or maybe there is some CSS that would allow me to show this new index like a table? I'm not UI expert, and with earlier Rails versions, I didn't needed to be.

I'm not the only one troubled with this:

https://www.reddit.com/r/rails/comments/pztx9g/scaffold_index_no_longer_generating_html_tables/
https://www.reddit.com/r/rails/comments/qypvnm/rails_new_scaffold_index/
https://stackoverflow.com/questions/71357307/no-table-view-after-crud-scaffolding
https://stackoverflow.com/questions/72402399/add-new-partial-to-rails-7-generate-scaffold

IMHO this change as it is, may be beneficial to some while detrimental to others. I feel like the small guy being left behind here.

I appreciate very much all the work you have done, @dhh along with all the Rails team.

Regards.

olivierlacan:
```
$ rails -v
Rails 7.1.3
rails new --css=tailwind rails-7-1-tailwind
cd rails-7-1-tailwind/
bin/rails g scaffold Post title:string description:text
bin/rails db:migrate
bin/rails s
```

Screenshot 2024-02-25 at 4 35 23 AM
Screenshot 2024-02-25 at 4 36 00 AM
Screenshot 2024-02-25 at 4 36 05 AM

@dixpac I'm curious what you did to obtain the output [shown in your post](https://github.com/rails/tailwindcss-rails/pull/71#issuecomment-929514917). Granted this was likely several versions behind what I used here. 

There are no browser console or Rails server errors indicating any issues with loading CSS. So is this is the expected stock output of Rails scaffolds in 7.1.3.2 with tailwindcss-rails 2.3.0?

Issue #77, opened 2021-09-30
`error tailwindcss [not found]` when generating a controller


yshmarov:
My setup:
```
rubRails 7.0.0.alpha2
ruby 3.0.1p64 (2021-04-05 revision 0fb782ee38) [x86_64-linux]
```
![image](https://user-images.githubusercontent.com/13472945/135510945-a0e914eb-6709-4a9f-8a8b-76a57853a97a.png)

yshmarov:
Also, when you hover on `submit`, mouse cursor does not change to `pointer`

![photo_2021-09-30_21-34-00](https://user-images.githubusercontent.com/13472945/135519064-93a31a02-9f45-4ced-9944-6ef1b9a64ff6.jpg)
![2](https://user-images.githubusercontent.com/13472945/135519061-f5f6039a-f1bc-4cc6-81aa-d0c3e63de4c4.jpg)

yshmarov:
Also it looks like `<h1>` tags don't work... other tags do

stuzero:
I have the same error error  `tailwindcss [not found]` when  generating a controller - the view is never created

```
ruby 3.0.2p107 (2021-07-07 revision 0db68f0233) [x86_64-linux]
Rails 6.1.4.1
```

rmontgomery429:
I am getting this error when generating [view_component](https://github.com/github/view_component). I saw that @dixpac just pushed a fix that was merged an hour ago and so I tried updating to main but, I'm still seeing the error.

I created a bare minimum reproduction rails app with only the [tailwindcss-rails](https://github.com/rails/tailwindcss-rails) and view_component [here](https://github.com/rmontgomery429/tailwindcss-rails-viewcomponent-conflict).

scott-knight:
@rmontgomery429 version 0.2.1 solves this issue

phanhaiquang:
I also got this issue when generating a "resource"

      create    app/controllers/<resource>_controller.rb
       error    tailwindcss [not found]

I am using 7.0.0.alpha2 
ruby 3.0.0p0 (2020-12-25 revision 95aff21468) [x86_64-darwin20]
tailwindcss-rails 0.5.0

PR #78, opened 2021-09-30
FIX: Error when generating a controller.

dixpac:
This PR fixes error on the controller generation:

`bin/rails g controller Home index`

Since, I'd to override rails scaffold controller to work with rails 6, I
needed to configure rails controller generator also.


dixpac:
Fixes the main issue of #77 . Other issues noted there, we will fix tomorrow since they are less pressing atm :)

dhh:
Can you fix up https://github.com/rails/tailwindcss-rails/issues/77?

PR #85, opened 2021-10-27
Close DIV generated for controller view

binhly:
The generated view for the controller was missing a closing </div>


dixpac (commented):
👍🏼  good catch

dixpac:
@dhh when you got some time can you merge this 😃

PR #89, opened 2021-11-14
Add support for Rails 5.

dixpac:
Giving this another go, since we need this in order to add tailwindcss generator for the `view_component`.

View component is `rails >= 5`

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,7 +2,7 @@ PATH
   remote: .
   specs:
     tailwindcss-rails (0.5.1)
-      rails (>= 6.0.0)
+      rails (>= 5.0.0)
 
 GEM
   remote: https://rubygems.org/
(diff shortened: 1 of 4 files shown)


dixpac:
@dhh I have opened the PR on view_components that fixes the issue with (`tailwindcss-rails <-> view_component`).
But, view component depends on rails >= 5.

So, if you think that supporting rails 5 makes sense, merge and deploy(🙏🏼 ) than I can finish the work on `view_component` PR 😄

dixpac:
@dhh pinging 😄

dhh:
I'm not really keen on support Rails 5. I don't think we need to support the same as view_component does. We can still make the two compatible.

dixpac:
Me neither. There is possibility atm to generate components wihtout the problem like this:
`bin/rails g component <Name> --template-engine=erb`

Main thing here https://github.com/github/view_component/pull/1098 is that view_component depends on rails 5 in their tests.

I'm closing this then, I will try to find the other way

PR #95, opened 2021-12-13
Upgrade to TailwindCSS v3

### LINKED ISSUE #94: Upgrade to Tailwind CSS v3
[Tailwind CSS v3.0 is out](https://tailwindcss.com/blog/tailwindcss-v3), so this issue can track the work (if any) needed for this gem to upgrade to it.

Their upgrade guide is [here](https://tailwindcss.com/docs/upgrade-guide).

coding-chimp:
This isn't really complete yet, but I wanted to see if this solution is acceptable before going down the rabbit hole too far.

As @dixpac mentioned in #94, Tailwind 3 removed classic mode and now only builds using JIT. This means there is no officially supported way to generate the whole CSS anymore. However, they do support safe listing classes with patterns. So I figure using the broadest pattern possible (`.+`) should pretty much safelist everything and give us the whole CSS again.

I still need to figure out a way to verify that there isn't anything missing for some reason and figure out which variants were included by default in v2. I'll do that once I get a thumbs up on the approach.

Resolves #94

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,11 +51,11 @@ By default, the CSS purger will only operate on the tailwind css file included w
 
 The Tailwind CSS main file that's being used before purging consists of these versions:
 
-* @tailwindcss/aspect-ratio 0.2.1
-* @tailwindcss/forms 0.3.3
-* @tailwindcss/typography 0.4.1
-* autoprefixer 10.3.1
-* tailwindcss 2.2.15
+* @tailwindcss/aspect-ratio 0.4.0
+* @tailwindcss/forms 0.4.0
+* @tailwindcss/typography 0.5.0
+* autoprefixer 10.4.0
+* tailwindcss 3.0.2
 
 
 ## License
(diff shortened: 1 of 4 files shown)


dixpac:
@coding-chimp, good idea, certainly only one possible atm.
Let me now when you finish the WIP, so I can test it :)

As for the variants only default for TW 2 are included https://v2.tailwindcss.com/docs/configuring-variants#default-variants-reference, but I guess we should include TW 3 default [variants](https://unpkg.com/browse/tailwindcss@3.0.2/stubs/defaultConfig.stub.js) if that doesn't increase the file size too much.

coding-chimp:
Yeah, I don't think this works like I want it to work. 😔 

I updated the config to only use variants enabled in v2 and am already getting a 10.4MB file. It basically generates all the base classes and then duplicates and prefixes _all of them_ for each variant. For one, this means the file is getting a little bloated, but even worse, it does not actually apply the styling associated with the variants. Most variants (hover, focus, etc.) don't have any special styling, but it does break the responsive variant.

dhh:
We can't support 3.0 with the current approach. The TW team has some ideas on how we can work around it, but we need help from their end.

PR #97, opened 2021-12-18
Remove redundant font-size class from generators

marcushwz:
Signed-off-by: Marcus Heng <marcushwz@gmail.com>

There are redundant class names for font-size in some of the default generators. This is what get generated when I do:

```bash
rails g controller static_pages home
```

image

In this PR, I have removed the `text-lg` from the affected generators. The comparison of `text-lg` and `text-4xl` are as follow:

- `text-4xl`
image

- `text-lg`
image


dixpac (commented):
👍❤️

Issue #98, opened 2021-12-18
Generator of main closing tag is broken if yield has arguments


martinadamek:
It happens for example in JumpstartRails templates, where one of yields uses is `<%= yield :title %>`.
Closing `</main>` tag is not inserted in application.html.erb and the file ends up broken after `./bin/rails tailwindcss:install` is executed.

dhh:
cc @dixpac

dixpac:
Thanks @dhh, I'll take a look tomorrow.

PR #106, opened 2021-12-21
Fix: Insert centring container around the yield

dixpac:
Insertion point of the centering container was problematic.
Meaning, if someone has a `yield :somethig`
inside theirs `application.html.erb` old regex would pick that as a
insertion point and possiblly break the html structure.

This commit makes that regex more rigid to only insert centering
container around the `<%= yield %>`.

Fixes: https://github.com/rails/tailwindcss-rails/issues/98

### DIFF
--- a/lib/install/tailwindcss.rb
+++ b/lib/install/tailwindcss.rb
@@ -1,12 +1,14 @@
-APPLICATION_LAYOUT_PATH = Rails.root.join("app/views/layouts/application.html.erb")
+APPLICATION_LAYOUT_PATH             = Rails.root.join("app/views/layouts/application.html.erb")
+CENTERING_CONTAINER_INSERTION_POINT = /^\s*<%= yield %>/.freeze
 
 if APPLICATION_LAYOUT_PATH.exist?
   say "Add Tailwindcss include tags and container element in application layout"
   insert_into_file APPLICATION_LAYOUT_PATH.to_s, <<~ERB.indent(4), before: /^\s*<%= stylesheet_link_tag/
     <%= stylesheet_link_tag "tailwind", "inter-font", "data-turbo-track": "reload" %>
   ERB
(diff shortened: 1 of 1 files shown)


lib/install/tailwindcss.rb
```diff
@@ -1,12 +1,14 @@
...
 
 if APPLICATION_LAYOUT_PATH.exist?
   say "Add Tailwindcss include tags and container element in application layout"
   insert_into_file APPLICATION_LAYOUT_PATH.to_s, <<~ERB.indent(4), before: /^\s*<%= stylesheet_link_tag/
     <%= stylesheet_link_tag "tailwind", "inter-font", "data-turbo-track": "reload" %>
   ERB
-  insert_into_file APPLICATION_LAYOUT_PATH.to_s, %(    <main class="container mx-auto mt-28 px-5 flex">\n  ), before: /^\s*<%= yield/
-  insert_into_file APPLICATION_LAYOUT_PATH.to_s, %(\n    </main>),  after: /^\s*<%= yield %>/
+
+  insert_into_file APPLICATION_LAYOUT_PATH.to_s, %(    <main class="container mx-auto mt-28 px-5 flex">\n  ), before: CENTERING_CONTAINER_INSERTION_POINT
```
dixpac, line 10: One more thing we could do here is to ask a question does user want to insert centering contianer
```ruby
if yes?("Do you want to insert self-centering html attribute inside your aplication.html.erb")
  insert_into_file APPLICATION_LAYOUT_PATH.to_s, %(    <main class="container mx-auto mt-28 px-5 flex">\n  ), before: /^\s*<%= yield %>/
  insert_into_file APPLICATION_LAYOUT_PATH.to_s, %(\n    </main>),  after: /^\s*<%= yield %>/
end
```

This would only be useful for folks installing `tailwindcss-rails` on already established app UI aesthetics

dixpac:
cc @dhh (not sure if you forgot about this patch, or you don't like the solution 😃 )

dhh:
All good 👍

PR #107, opened 2021-12-21
Docs: Add note about `view_component`

dixpac:
This is the way how to generate `view_component` successfully atm.

---------------

Current state here is that I've opened the PR to `view_component` to
support the `tailwindcss-rails` generators.
But, in order to merge that PR they are in the process of introducing the `appraisal` gem, so they
can test against multiple dependecy versions.

Until that is done, this is the way how to generate a component.


dixpac:
This is no longer necessary since view_component now works with tailwindcss-rails https://github.com/github/view_component/pull/1098

PR #117, opened 2021-12-28
Support for arm64-darwin-21

juanmanuelramallo:
## Steps to reproduce
1. Make sure to run these commands on a M1 Pro machine
2. Run `./bin/bundle add tailwindcss-rails`
3. Run `./bin/rails tailwindcss:install` 

## Expected
Tailwind is correctly installed in the app

## Actual
The following error is displayed:
```
ERROR: Cannot find the tailwindcss executable for arm64-darwin-21 in ~/.gem/ruby/3.0.3/gems/tailwindcss-rails-2.0.2/exe
If you're using bundler, please make sure you're on the latest bundler version:

  gem install bundler
  bundle update --bundler

Then make sure your lock file includes this platform by running:

  bundle lock --add-platform arm64-darwin-21
  bundle install

See `bundle lock --help` output for details.
```

## Description
Apple's M1 Pro chip is referenced as arm64-darwin-21. So the
installation process can't succeed because the executable file is
missing for the platform.

In order to test this, I did the following:
1. Cloned the repo,
2. Updated the NATIVE_PLATOFORMS constant
3. Ran `rake gem:arm64-darwin-21`
4. Copied the newly generated folder from the gem's repo `./exe/arm64-darwin-21` to the gem's folder in my machine `~/.gem/ruby/3.0.3/gems/tailwindcss-rails-2.0.2/exe/arm64-darwin-21`
5. Ran the `./bin/rails tailwindcss:install` command and everything worked successfully
6. Tailwind was properly installed

Perhaps we could update the package.rake to alias arm64-darwin to arm64-darwin-21 or something similar. Perhaps this will make more sense if it becomes a trend for Apple to change their processor's arch name after a new release. I'm still unsure if the M1 Max has a different name, I'd bet it does.

PR #122, opened 2022-01-04
Add additional recovery step when running into platform related issues.

andrewculver:
Not sure the ideal wording here, but after going on the journey recorded for all posterity in https://github.com/rails/tailwindcss-rails/issues/121, I thought maybe it would make sense to add some instructions to avoid someone else suffering the same fate in the future.

### DIFF
--- a/exe/tailwindcss
+++ b/exe/tailwindcss
@@ -34,7 +34,9 @@ if exe_path.nil?
     See `bundle lock --help` output for details.
         
     If you're still seeing this message after taking those steps, try running
-    `bundle config` and ensure `force_ruby_platform` isn't set to `true`.
+    `bundle config` and ensure `force_ruby_platform` isn't set to `true`. See
+    https://github.com/rails/tailwindcss-rails#check-bundle_force_ruby_platform
+    for more details.
   ERRMSG
   exit 1
 end
--- a/exe/tailwindcss
+++ b/exe/tailwindcss
@@ -32,6 +32,9 @@ if exe_path.nil?
       bundle install
 
     See `bundle lock --help` output for details.
+        
+    If you're still seeing this message after taking those steps, check that you 
+    don't have `BUNDLE_FORCE_RUBY_PLATFORM` set to `true` in `~/.bundle/config`.
   ERRMSG
   exit 1
 end
--- a/exe/tailwindcss
+++ b/exe/tailwindcss
@@ -33,8 +33,8 @@ if exe_path.nil?
 
     See `bundle lock --help` output for details.
         
-    If you're still seeing this message after taking those steps, check that you 
-    don't have `BUNDLE_FORCE_RUBY_PLATFORM` set to `true` in `~/.bundle/config`.
+    If you're still seeing this message after taking those steps, try running
+    `bundle config` and ensure `force_ruby_platform` isn't set to `true`.
   ERRMSG
   exit 1
 end


flavorjones:
I'm in favor of this change, since #121 and #117 provide evidence that multiple people are being confused by this bundler config setting.

flavorjones:
Note I've also submitted a PR to add a "troubleshooting" section to the README that informs users about this edge case: #123

andrewculver:
@flavorjones Thanks for adding it to the docs! I think it still makes sense to have a message like what's being proposed in this PR, because folks running into this issue via `rails new example-application --css=tailwind` may not even know about this specific Ruby Gem. Do you think we should add a link to your troubleshooting guidance in https://github.com/rails/tailwindcss-rails#check-bundle_force_ruby_platform from the error message?

flavorjones:
@andrewculver Linking to the troubleshooting section is a good idea! Would you mind updating this PR to do that? Then we can poke DHH to merge.

andrewculver:
@flavorjones Done!

flavorjones:
@SampsonCrowley Thanks for asking this question. There are a few variations on packaging we could have chosen, but each has its own tradeoffs.

Generally it's good for a gem package to contain all the artifacts it needs, and not rely on an external network connection (not all systems will have access at install time for security reasons) or that relies on the remote repository being up and online and not throttling requestors. I've personally had bad experiences as a maintainer when Nokogiri relied on downloading source libraries for zlib, libxml, etc. at install time (circa 2009). Throttling in particular was very hard for users to work around and for me to help them with.

We could also have packaged _all_ the binaries into a single universal gem, but then the gem would be quite large and that has drawbacks of its own (network costs for rubygems.org, slower installation, more diskspace taken up in an installation).

The current set of tradeoffs is: small, single-architecture rubygems with a tradeoff of requiring an extra step for a Gemfile.lock to work across multiple architectures. Generally you can do this by updating to a modern version of bundler and running `bundle lock --add-platform=<platform>` for each platform (dev, prod, whatever). Committing that info to the lockfile means you shouldn't need to do it again after that.

I'm not sure I understand the problem you're having when you say "kills generic gem vendor caching". If you are experiencing problems, can I ask you to open a new issue to discuss how to solve them?

SampsonCrowley:
@flavorjones thanks for the quick reply. Before I open a new issue the general problem that people who tend to use force_ruby_platform are trying to solve is that there is zero way to bundle for other platforms in gem caching/secure install setups. In the company I'm with, we're loading and installing packages of gems from the same repo probably 200x a day on the low end, for only one single repo. So to minimize resources and potential security leakes, larger apps and apps that require higher security utilize a code-owned vendor cache. So bundler in production, for security reasons is set to only install from vendor cache. But increasingly, for alternate IT related security reasons, all dev machines are only being allowed to be OSx.

So in order to add the gems that deployments can build that are reviewed and vetted, the few of us that do have access have to drop into a docker image, and bundle again on Linux platform purely for this gem because there is no "ruby" platform option.

Not all devs completely understand what it takes to manage multiple platforms in a vendor cache and to be sure to always add Linux-target cache gems even when developing on osx, and devs who are on Linux have no way to add Darwin gems so when OSx devs pull new gems, that adds additional git churn

"ruby" platform option is cross platform, requires no additional deployment management and has the same content across platforms.

Mostly just want to note this as docs for the future, as I don't know what the solution is to the network issues you mentioned other than specifically gitignoring all Darwin gems, and that leaves a huge network overhead for development. But it would be nice to remove that management overhead one day in the future for people who do need the `vendor/cache` on larger multi-platform developer projects

flavorjones:
I wonder if the bundler team would be open to considering a feature like `--cache-all-platforms` to help manage multiple platforms for teams like yours? This is just one native gem, there are others (sqlite3, nokogiri, rubymagick just to name a few) and I don't think we can fix this problem by shifting the burden onto each gem maintainer.

PR #131, opened 2022-01-09
 Remove setting of tailwindcss as the default template_engine

davidteren:
Not sure why we set `g.template_engine :tailwindcss` it is not a typical template engine like `erb, haml, slim` and others.

One adverse effect of setting it to `tailwindcss` is RSpec view generators will  not use the correct template name.
Eg: `spec/views/home/index.html.tailwindcss_spec.rb`  where it should be `spec/views/home/index.html.erb.rb`

Removing the config resolves the above issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,28 @@ With Rails 7 you can generate a new application preconfigured with Tailwind by u
 1. Run `./bin/bundle add tailwindcss-rails`
 2. Run `./bin/rails tailwindcss:install`
 
+## Using the included TailwindCSS generator templates
+
+Included are several pre-styled templates. 
+
+```
+Tailwindcss:
+  tailwindcss:controller
+  tailwindcss:mailer
+  tailwindcss:scaffold
+ ```
+
+Simply prepend `tailwindcss:` to the generator type for some styling.
+
+```
+rails generate tailwindcss:scaffold Post title:string content:text 
+```
+These templated styles have been set up with a container element in mind. This element is added to `views/layouts/application.html.erb` during the above install process.
+```
+<main class="container mx-auto mt-28 px-5 flex">
+  <%= yield %>
+</main>
+```
 
 ## Building in production
 
--- a/lib/tailwindcss/engine.rb
+++ b/lib/tailwindcss/engine.rb
@@ -9,9 +9,5 @@ module Tailwindcss
     initializer "tailwindcss.disable_generator_stylesheets" do
       Rails.application.config.generators.stylesheets = false
     end
-
-     config.app_generators do |g|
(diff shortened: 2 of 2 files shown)


dixpac:
Yeh, this will break the scaffolding, if we remove the `template_engine` we'll not get tailwind scaffolds. This is done in order to override the default rails scaffolds https://github.com/rails/tailwindcss-rails/tree/main/lib/generators/tailwindcss.

I'm 100% up for removing the `template_engine` if you can find the way to override rails scaffolds so we have `_partial.html.erb` for rails 6 versions

dixpac:
This will break the scaffolding afaik, if we remove the template_engine we'll not get tailwind scaffolds. This is done in order to override the default rails scaffolds https://github.com/rails/tailwindcss-rails/tree/main/lib/generators/tailwindcss.

I'm 100% up for removing the template_engine if you can find the way to override rails scaffolds so we have `_partial.html.erb` for rails 6 versions.

davidteren:
Thanks for the insights. I'll check properly if the scaffolds work with the configuration removed.

davidteren:
@dixpac turns out none of the scaffolds will include the tailwind custom styling with my change. This may be a good thing as this no longer forces it on all users.
If a user wants the custom styling that comes with the gem they can simply prepend the styling directive to the generator.
See the updated Readme.

dixpac:
Thanks @davidteren 

I don't think this is the direction we wan't to go with, since our main idea with the tailwind generators was to override default rails generators so we don't have to provide `tailwindcss:scaffold`.

@dhh has the last word on this, but the initial idea was just to use `bin/rails g scaffold ...` and have the Tailwind styles if this engine is included.

dixpac:
[Reference](https://github.com/rails/tailwindcss-rails/issues/23#issuecomment-773344368)

davidteren:
@dixpac I think we can close this PR as in its current form it does not align with @dhh's vision of having scaffold templates that demonstrate the use of Tailwind classes.

dixpac:
Yup!
But, if you have an inspiration or ideas in the future on how to keep default templates and fix these edge cases.. 🙏🏼

marckohlbrugge:
FWIW, this breaks the `haml-rails` default configuration and presumably similar integrations.

I didn't expect including the `tailwindcss-rails` gem would switch my generated views back from `.html.haml` to `.html.erb`

dhh:
Feel free to explore ways where all this can coexist. But yes, with the tailwind gem installed, scaffolds should be generated with TW styling. Maybe we can detect if another, non-default template handler has already been registered, and then warn/abort, though.

Issue #137, opened 2022-01-31
Using images from the asset pipeline


atimmer:
It would be useful for a Tailwind + Rails integration to include a way to use images from the asset pipeline. As far as I know it's not possible to use `asset-url` in any of the locations relevant to tailwind:

1. **In the `application.tailwind.css`** as a custom utility: This half-works, as `asset-url` is put in the `tailwind.css` output file. However, the asset pipeline ignores it because it isn't processed. Maybe I can enable this somehow?
2. **In the html.erb file** as an ['arbitrary value'](https://tailwindcss.com/docs/background-image#arbitrary-values). This doesn't work because Tailwind classes need to be hardcoded for automatic purposes.
3. **In the tailwind.config.js**. Doesn't work because the file has no relation to the asset pipeline at all.

I feel like 1. should be the 'intended' solution. However, I don't have enough knowledge of neither Rails nor the asset pipeline to write a PR for this. If someone provides me with pointers I can try creating a PR that makes it work.

There are two workarounds. The first is putting your images in the `public/assets` folder. However that means throwing away the advantages of the asset pipeline, including fingerprinting of the images allowing long cache times. The second is maintaining a separate CSS file that is solely responsible for images. That's not ideal, because it's confusing why you would have two CSS outputs even though you're using Tailwind. And you can't use Tailwind's `after:`, `before:`, `hover:`, etc.

StackOverflow question asking the same thing: https://stackoverflow.com/questions/66842651/inserting-a-background-image-from-asset-pipeline-with-tailwind.

dhh:
`url()` is rewritten automatically since https://github.com/rails/sprockets-rails/pull/476 👍

atimmer:
Awesome! I'll definitely answer the StackOverflow question with the solution. Do you think this should also be documented in the README of this project? Or is this something that will eventually be 'standard practice'?

dhh:
I'd be happy seeing a note in the README about it 👍

PR #140, opened 2022-02-05
Add a note about using asset pipeline assets

atimmer:
See #137

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,27 @@ A fix for this has been proposed upstream at https://github.com/tailwindlabs/tai
 apk add build-base gcompat
 ```
 
+### Using asset-pipeline assets
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -90,6 +90,27 @@ A fix for this has been proposed upstream at https://github.com/tailwindlabs/tai
 apk add build-base gcompat
 ```
 
+### Using asset-pipeline assets
+
+To use assets from the asset pipeline one would normally use `asset-url`. This doesn't work with Tailwind, as the CSS is not processed with sass. Instead use `url(image.svg)`, [since Sprockets v3.3.0](https://github.com/rails/sprockets-rails/pull/476) the URL will automatically be rewritten. In Tailwind this looks like this:
```
dixpac, line 95: `the URL will automatically be rewritten` for me as a reader this is a bit too much abstract, can we expand a bit on this so I don't have to open PR link to understand what rewritten means.

dixpac (commented):
Looks good to me, I've just left one small reamark.
But, this is definitely helpful 👍🏼 

cc @dhh

dhh:
We should just assert how it works. No need to confuse it with what used to be true in the past.

atimmer:
I have rewritten the paragraph that explains how it works.

PR #146, opened 2022-02-20
Installer: don't add main in existing projects

### LINKED ISSUE #139: Don't add main element to application.html.erb
When installing Tailwind it adds a `main` element like this `<main class="container mx-auto mt-28 px-5 flex">` to `application.html.erb`. While this might be fine with a newly created apps, it's annoying for existing apps since it messes with your existing layout.

I'd either remove this completely or only add it if the `<body>` of `application.html.erb` looks like a default Rails app.

I'm happy to look into this.

shafy:
This addresses and closes #139.

Currently, the installer adds a `<main>` element around `<%= yield %>` in `application.html.erb`. This is ok for new projects, but it's annoying for projects that add this gem later on since they likely already have changed their `application.html.erb` and installing the gem breaks the layout.

This proposed change first checks if `application.html.erb` looks standard around `<%= yield %>`, and only in that case wraps it in a `<main>` element.

### DIFF
--- a/lib/install/tailwindcss.rb
+++ b/lib/install/tailwindcss.rb
@@ -7,8 +7,10 @@ if APPLICATION_LAYOUT_PATH.exist?
     <%= stylesheet_link_tag "tailwind", "inter-font", "data-turbo-track": "reload" %>
   ERB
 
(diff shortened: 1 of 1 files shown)


dixpac:
I'm ok with this approach. But @dhh has the last word.

Issue #147, opened 2022-02-21
Provided asset file does not match upstream version


brenogazzola:
While working on https://github.com/rails/propshaft/issues/74 I've noticed that the file `app/assets/stylesheets/tailwind.css` is for version `2.2.19`, but the version in `upstream.rb` is for `3.0.22`.

Is the asset file still needed? If yes, shouldn't it be for the same version defined `upstream.rb`? This is causing some confusion in Propshaft, as it automatically grabs gem files inside `app/assets/stylesheets`.

HusseinMorsy:
I have the same issue with the combination of propshaft (0.6.3) and tailwind. I have therefore created PR #148

tylerdiaz:
Had the same issue using propshaft + tailwind, @HusseinMorsy's fix does the trick.

dixpac:
@brenogazzola it is not needed, nice catch. This PR should fix the issue https://github.com/rails/tailwindcss-rails/pull/148

PR #148, opened 2022-02-21
Delete legacy tailwind.css

HusseinMorsy:
As described in issue #147, I also suggest to remove the legacy tailwind.css which includes tailwindcss 2.2.19.


dixpac:
👍🏼  thanks!

PR #155, opened 2022-03-09
Only include ERB templates by default

pixeltrix:
Using `./app/views/**/*` includes transient files like vim backup files and generates errors like that seen in rails/rails#44646.

Obviously if people start using other template engines like Slim or Haml then they'll need to add the appropriate extensions.


dhh:
I think we changed this at one point to include everything to get haml files also? Maybe we could expand it to include haml and slim or whatever is reasonably popular. But 👍 on not having it be everything.

pixeltrix:
Added Slim, Haml and HTML since Rails can render raw HTML as well

PR #163, opened 2022-04-10
Move minify flag only for build task

achmiral:


dhh:
What's the purpose of not just keeping it consistent?

achmiral:
It's a little bit hard to debug when the `--minify` flag is on. For example when the styling is not applied on some element. So I have to check whether the class or styling has been generated or not in chrome developer tools or in "tailwind.css", especially for custom classes defined using the `@layer component` directive.

ur5us:
To add another data point: I’m combining Bootstrap v3 with Tailwind in app. Admittedly, just is just a transition period. Anyway, Bootstrap v3 and many other CSS frameworks make their own assumptions by using a particular CSS reset for instance which usually includes CSS declarations for `html { … }` and `body { … }`. In those case, debugging with a minified Tailwind build is tricky.

flavorjones:
I've create #198  which implements this feature more flexibly.

flavorjones:
v2.0.13 has this feature, please see [the README](https://github.com/rails/tailwindcss-rails#update-assets-automatically) for an explanation.

PR #168, opened 2022-04-25
Enable polling support for tailwindcss cli

thebub:
### Reasoning
Due to an WSL2 issue (https://github.com/microsoft/WSL/issues/4739) the watch mechanism of the tailwindcss cli will not work on all potential developer setups. Especially when using vscode-devcontainers and keeping the code in the mounted directory. (assumed default behavior)
The recommended work around is to use polling instead of the normal inotify based watch mechanism.

### Changes
* This pull request will add a new task to tailwindcss-rails to use the polling option instead of the watch option
* A parameter is added to the ./bin/dev script, to switch between watch and poll mechanism
* Default behavior of tailwindcss-rails remains unchanged

### DIFF
--- a/lib/install/Procfile.dev
+++ b/lib/install/Procfile.dev
@@ -1,2 +1,3 @@
 web: bin/rails server -p 3000
 css: bin/rails tailwindcss:watch
+css_polling: bin/rails tailwindcss:poll
--- a/lib/install/dev
+++ b/lib/install/dev
@@ -6,4 +6,18 @@ then
   gem install foreman
 fi
 
-foreman start -f Procfile.dev
+while getopts p: option
+do 
+  case "${option}"
+    in
+    p)polling=1;;
+    *);;
+  esac
+done
+
+if [ $polling -eq 1 ]
+then
+  foreman start -f Procfile.dev -m all=1,css=0
+else 
+  foreman start -f Procfile.dev -m all=1,css_polling=0
+fi
--- a/lib/install/dev
+++ b/lib/install/dev
@@ -6,7 +6,7 @@ then
   gem install foreman
 fi
 
-while getopts p: option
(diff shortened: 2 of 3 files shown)


thebub:
Waiting for TailwindCSS to ship the polling feature. The feature was omitted in the recent v3.0.24 release.

daveomcd:
@thebub I have a new rails 7 application with tailwindcss installed.  None of my CSS changes are reflected when reloading my pages.  Is this a current limitation to working with WSL2?  Or perhaps these issues are separate. Thanks.

hyphenized:
@daveomcd I believe this is related to the wsl2 problem mentioned before, try moving your project files to your linux home folder (e.g. `/home/daveomcd` or `~`) and see if the problem still happens

thebub:
@daveomcd I suggest you follow the recommendation by @hyphenized and provide feedback, whether the problem is resolved or not.

daveomcd:
> @daveomcd I suggest you follow the recommendation by @hyphenized and provide feedback, whether the problem is resolved or not.

Hey @thebub, thank you for the response.  I figured out the issue I was seeing wasn't a problem with WSL2.  It was because I ran `rake assets:precompile` by accident.  To correct the problem I had I just needed to delete all the files in my public/assets folder.

flavorjones:
I've created #199 which uses a task argument to implement this, which is a bit more flexible than a separate task (and can be combined with other options like `debug` from #198).

flavorjones:
v2.0.13 has this feature, please see [the README](https://github.com/rails/tailwindcss-rails#update-assets-automatically) for an explanation.

thebub:
Thanks @flavorjones. I wasn't aware of the possibility to add options as well. This is a way better solution!

PR #170, opened 2022-05-03
Docs: Add notes for running tailwindcss:watch in Docker

geetfun:
While running `bin/rails tailwindcss:watch` from within a `Procfile`, it was noticed that the command returns an exit 0 status if this is happening from within a Docker container.

See https://github.com/tailwindlabs/tailwindcss/issues/5324 for same problem and solution.

This may present as a gotcha for those running their Rails app from `docker-compose.yml`. Suggestion here to add this useful piece of information in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ The installer will create your Tailwind input file in `app/assets/stylesheets/ap
 
 If you need to use a custom input or output file, you can run `bundle exec tailwindcss` to access the platform-specific executable, and give it your own build options.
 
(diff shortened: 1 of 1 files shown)


paddy-hack:
Ran into this myself and found the same issue you mentioned :sweat_smile:

dixpac (commented):
This is helpful. But, maybe we don't need new sub-heading for this can we just add note about docker in the beginning where `tailwindcss:watch` is mentioned.

Somewhere around this paragraph:

> When you're developing your application, you want to run Tailwind in watch mode, so changes are automatically reflected in the generated CSS output. You can do this either by running rails tailwindcss:watch as a separate process, or by running ./bin/dev which uses [foreman](https://github.com/ddollar/foreman) to starts both the Tailwind watch process and the rails server in development mode.

geetfun:
> This is helpful 👍🏼 . But, maybe we don't need new sub-heading for this can we just add note about docker in the beginning where `tailwindcss:watch` is mentioned.
> 

Good idea. I've moved it to your suggested paragraph where `rails tailwindcss:watch` is mentioned.

PR #179, opened 2022-06-23
Make config path configurable

samuel02:
I ran into an issue where I had to add `type: "module"` to my `package.json`. This led to an error running Tailwind:

```
Error [ERR_REQUIRE_ESM]: require() of ES Module /Users/samuelnilsson/projects/plat_nisse/config/tailwind.config.js from /snapshot/tailwindcss/lib/cli.js not supported.
tailwind.config.js is treated as an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which declares all .js files in that package scope as ES modules.
Instead rename tailwind.config.js to end in .cjs, change the requiring code to use dynamic import() which is available in all CommonJS modules, or change "type": "module" to "type": "commonjs" in /Users/samuelnilsson/projects/plat_nisse/package.json to treat all .js files as CommonJS (using .mjs for all ES modules instead).
```

So, naturally, I renamed my config to end in .cjs but then I ran into:

```
Specified config file [REDACTED]/config/tailwind.config.js does not exist.
```

I found out this gem hardcodes the location of the config file and added the ability to use a custom path via an ENV variable. What do you think?

BTW, when testing my solution I pointed the gem to my fork on GitHub but ran into issues since it didn't contain the binaries. Is there a good way of solving that except for having to publish my own fork? I just went to the gem location and ran `rake package` locally but that won't fly in production.

### DIFF
--- a/lib/tasks/build.rake
+++ b/lib/tasks/build.rake
@@ -1,4 +1,9 @@
-TAILWIND_COMPILE_COMMAND = "#{RbConfig.ruby} #{Pathname.new(__dir__).to_s}/../../exe/tailwindcss -i '#{Rails.root.join("app/assets/stylesheets/application.tailwind.css")}' -o '#{Rails.root.join("app/assets/builds/tailwind.css")}' -c '#{Rails.root.join("config/tailwind.config.js")}' --minify"
+TAILWIND_COMPILE_COMMAND = "#{RbConfig.ruby} "\
+  "#{Pathname.new(__dir__)}/../../exe/tailwindcss "\
+  "-i '#{Rails.root.join("app/assets/stylesheets/application.tailwind.css")}' "\
+  "-o '#{Rails.root.join("app/assets/builds/tailwind.css")}' "\
+  "-c '#{Rails.root.join(ENV.fetch("TAILWIND_CONFIG_PATH", "config/tailwind.config.js"))}' "\
+  "--minify"
 
 namespace :tailwindcss do
   desc "Build your Tailwind CSS"


dhh:
If you're using package.json, and thus node in general, you don't need this gem. This is only intended for node-less setups.

samuel02:
@dhh I see where you are coming from but I disagree with your conclusion. I have a "vanilla" Rails 7 app, and want to keep it as vanilla as possible, but need to execute a node script (from a background job). This node script uses some code from the app which naturally uses ES6 modules. So in order for the node script to use the the code, package.json needs to have `"type": "module"` and then I'm down the rabbit hole I described before. But for now I will just re-implement the rake task:

```ruby
Kernel.silence_warnings do
  TAILWIND_COMPILE_COMMAND = "#{RbConfig.ruby} "\
    "#{Bundler.rubygems.find_name("tailwindcss-rails").first.full_gem_path}/exe/tailwindcss "\
    "-i '#{Rails.root.join("app/assets/stylesheets/application.tailwind.css")}' "\
    "-o '#{Rails.root.join("app/assets/builds/tailwind.css")}' "\
    "-c '#{Rails.root.join("config/tailwind.config.cjs")}' "\
    "--minify"
end

namespace :tailwindcss do
  desc "Build your Tailwind CSS"
  task :build do
    system TAILWIND_COMPILE_COMMAND
  end

  desc "Watch and build your Tailwind CSS on file changes"
  task :watch do
    system "#{TAILWIND_COMPILE_COMMAND} -w"
  end
end

Rake::Task["assets:precompile"].enhance(["tailwindcss:build"])

if Rake::Task.task_defined?("test:prepare")
  Rake::Task["test:prepare"].enhance(["tailwindcss:build"])
elsif Rake::Task.task_defined?("db:test:prepare")
  Rake::Task["db:test:prepare"].enhance(["tailwindcss:build"])
end
```

I won't fight for this feature though, I will just leave this comment here if anyone else encounters the same issue.

PR #182, opened 2022-07-07
Tailwind plugins

aaripurna:
This PR tries to add support for tailwind UI plugins. So the client only needs to define it in `config/tailwind.plugins.yml`
```yaml
# config/tailwind.plugins.yml

# Format
# plugin_name: cdn URI
# Example
daisyui: https://cdn.jsdelivr.net/npm/daisyui@2.18.1/dist/full.css
```

Then add the plugin inside `config/tailwind.config.js`
```javascript
// other code

module.export = {
  // other code
  plugins: [
    // other code
    usePlugin('daisyui')
  ]
}
```


dhh:
Appreciate the effort here, but this is too much custom logic for what this gem is trying to do. We try to keep it to the absolute minimum, and only what's included with TW by default. But feel free to add this as a standalone extension.

PR #186, opened 2022-08-08
Update to Tailwind CSS v3.1.8

TastyPi:
I need the fix in https://github.com/tailwindlabs/tailwindcss/pull/8687 which was released in v3.1.4.

Has there been any thought in decoupling `tailwindcss-rails` releases from Tailwind? It would be nice to be able to use the latest Tailwind release without having to wait for `tailwindcss-rails` to be released.

### DIFF
--- a/lib/tailwindcss/upstream.rb
+++ b/lib/tailwindcss/upstream.rb
@@ -1,7 +1,7 @@
 module Tailwindcss
   # constants describing the upstream tailwindcss project
   module Upstream
-    VERSION = "v3.1.3"
+    VERSION = "v3.1.8"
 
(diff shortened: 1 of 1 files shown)


flavorjones (approved):
LGTM

flavorjones:
@TastyPi Thanks for submitting this. You're asking a good question about decoupling, and I think we're optimizing for the user experience over latest-and-greatest version with the current setup.

Would you mind opening a new issue to start a conversation about supporting folks who have tailwindcss already installed on their system?

TastyPi:
@flavorjones Thanks for the review, I've opened #187 to continue that discussion.

PR #189, opened 2022-08-09
address issue when running commands with #exec

blerchin:
This is a followup to https://github.com/rails/tailwindcss-rails/pull/181

After upgrading one of our repos to `2.0.11`, we unfortunately learned that running commands with `Kernel#exec` instead of `Kernel#system` introduces a failure in the `assets:precompile` pipeline. Due to `exec` hijacking the process from Ruby, subsequent steps will not run after the tailwind build completes.

This will probably need to be treated as an urgent bugfix as it may result in a failure of the asset pipeline to output images, javascript, etc.

This change effectively reverts #181, and adds the `exception: true` flag which will allow `#system` to raise when encountering failure codes.

### DIFF
--- a/lib/tasks/build.rake
+++ b/lib/tasks/build.rake
@@ -3,12 +3,12 @@ TAILWIND_COMPILE_COMMAND = "#{RbConfig.ruby} #{Pathname.new(__dir__).to_s}/../..
 namespace :tailwindcss do
   desc "Build your Tailwind CSS"
   task :build do
-    exec TAILWIND_COMPILE_COMMAND
+    system(TAILWIND_COMPILE_COMMAND, exception: true)
   end
 
   desc "Watch and build your Tailwind CSS on file changes"
   task :watch do
-    exec "#{TAILWIND_COMPILE_COMMAND} -w"
+    system "#{TAILWIND_COMPILE_COMMAND} -w"
   end
 end
 


masterkain:
just got hit by this

philipithomas:
Also affected by this 👍🏻

lib/tasks/build.rake
```diff
@@ -3,12 +3,12 @@ TAILWIND_COMPILE_COMMAND = "#{RbConfig.ruby} #{Pathname.new(__dir__).to_s}/../..
...
   desc "Build your Tailwind CSS"
   task :build do
-    exec TAILWIND_COMPILE_COMMAND
+    system(TAILWIND_COMPILE_COMMAND, exception: true)
   end
 
   desc "Watch and build your Tailwind CSS on file changes"
   task :watch do
-    exec "#{TAILWIND_COMPILE_COMMAND} -w"
+    system "#{TAILWIND_COMPILE_COMMAND} -w"
```
jacob-carlborg-apoex, line 11: Is there a reason `exception: true` is not used here? Even though the command is not expected to terminate, it can still fail to start in the first place.

flavorjones: No reason other than it hasn't been a problem we needed to solve. If you'd like to submit a PR I'd consider merging it.

PR #194, opened 2022-09-01
fix: ship an x86_64-linux-musl platform gem

flavorjones:
This was not strictly necessary before now because musl platforms have
always used the generic platform gem, but recent changes in rubygems
and bundler introduced new behaviors around libc flavors, and so we're
going the extra mile here to be explicit.

See #192 for context, also see
https://github.com/rubygems/rubygems/pull/5875

### DIFF
--- a/.github/workflows/gem-install.yml
+++ b/.github/workflows/gem-install.yml
@@ -5,7 +5,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        platform: ["ruby", "x64-mingw32", "x64-mingw-ucrt", "x86_64-darwin", "x86_64-linux"]
+        platform: ["ruby", "x64-mingw32", "x64-mingw-ucrt", "x86_64-darwin", "x86_64-linux", "x86_64-linux-musl"]
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
@@ -45,6 +45,19 @@ jobs:
       - uses: actions/download-artifact@v2
         with:
           name: gem-x86_64-linux
+          path: pkg
+      - run: "gem install pkg/tailwindcss-rails-*.gem"
+      - run: "tailwindcss --help"
+
+  linux-musl-install:
+    needs: ["package"]
+    runs-on: ubuntu-latest
+    container:
+      image: ruby:3.1-alpine
+    steps:
+      - uses: actions/download-artifact@v2
+        with:
+          name: gem-x86_64-linux-musl
           path: pkg
       - run: "gem install pkg/tailwindcss-rails-*.gem"
       - run: "tailwindcss --help"
--- a/README.md
+++ b/README.md
@@ -50,6 +50,7 @@ Some users are reporting this error even when running on one of the supported na
 - x64-mingw-ucrt
 - x86_64-darwin
 - x86_64-linux
+- x86_64-linux-musl
 - aarch64-linux
 
 #### Check Bundler PLATFORMS
--- a/lib/tailwindcss/upstream.rb
+++ b/lib/tailwindcss/upstream.rb
@@ -10,6 +10,7 @@ module Tailwindcss
       "x64-mingw-ucrt" => "tailwindcss-windows-x64.exe",
       "x86_64-darwin" => "tailwindcss-macos-x64",
       "x86_64-linux" => "tailwindcss-linux-x64",
+      "x86_64-linux-musl" => "tailwindcss-linux-x64",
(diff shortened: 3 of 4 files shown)


flavorjones:
I'm going to wait a bit to hear back from rubygems/bundler maintainers that this is the appropriate thing to do here. Please don't merge yet.

flavorjones:
OK, seems like the bundler fix isn't going to ship for another week.

I suspect this is low-priority for them because the _majority_ of native platform gems can "fall back" safely to the vanilla platform (with a compilation penalty of some kind). This gem unfortunately cannot fall back, so ...

I think we should merge this, and either

1. ship a patch release of all the platforms,
2. or else run this updated `rake package` task against the `v2.0.12` tag and ship a `x86_64-linux-musl` platform gem for 2.0.12.

@dhh do you have a preference? I have a slight preference for **1** and will move forward with that plan later today unless you present objections.

flavorjones:
Hey @deivid-rodriguez, that's correct, I can only reproduce the issue with rubygems 3.3.21 on musl (Alpine).

I'm not confident this change actually makes things better for most people ... the issue I can most easily reproduce is the "looping" issue described at https://github.com/rubygems/rubygems/issues/5871

Will this change fix that looping during bundler resolution?

Can you suggest anything else we can do to work around this until an upstream fix is released?

deivid-rodriguez:
I think this PR should fix the looping, yes. But it may cause musl variants to get installed on glibc linux for users of old RubyGems. If the contents of the gems are exactly the same and you're just releasing them to workaround this issue, then I guess it doesn't matter although it's going to surprise people and feels wrong.

I don't have a good suggestion for you as `tailwindcss-rails` maintainers, but users can workaround easily by sticking to the previous RubyGems version. If you really feel this is urgent, I can try to move the release a bit earlier.

deivid-rodriguez:
> I tested this (installed the musl gem generated from this PR's branch on a glibc system) and it works OK.

I think it's pretty much unreliable. Sometimes it will install the correct variant, sometimes it won't.

> I don't feel strongly that we need to urgently address this -- we've only had one bug report, and there is a workaround ... though that may suddenly get more urgent if the default ruby:3.1-alpine image gets updated with the offending rubygems version 😆 😭 .

They don't do that anymore, actually. The alpine image comes with the default RubyGems version that comes with each Ruby.

deivid-rodriguez:
> I tested this (installed the musl gem generated from this PR's branch on a glibc system) and it works OK.

Sorry, I misunderstood this and my previous reply makes no sense. I see now, the gem would still work ok if the RubyGems bug was triggered and the musl variant got installed on non-musl platforms. Still feels wrong to me to publish separate variants if both work equally on both platforms.

woodhull:
for what it's worth, I also experienced this. Installing the latest rubygems fixes another problem I had, but then it causes this defect.

raybradley:
> I don't feel strongly that we need to urgently address this -- we've only had one bug report, and there _is_ a workaround ...
Sorry to be dense but I don't understand the workaround. I'm experiencing the problem on ruby:3.0.2-alpine. Never needed to roll back rubygems version...any pointers here?

flavorjones:
@raybradley the `ruby:3.0.2-alpine` image comes with a compatible version of rubygems and bundler by default:

```
$ docker run -it ruby:3.0.2-alpine /bin/sh
/ # gem -v
3.2.22
/ # bundler -v
Bundler version 2.2.22
```

Deivid's comment is saying "don't install the new version of rubygems or bundler if you're affected by this issue", not "roll back". Though, I do think he's saying that if your image installs rubygems 3.3.21 or bundler 2.3.21 and you are affected by this issue, you should update your image to not use those versions for now.

deivid-rodriguez:
Yes, sorry if I was not clear. What @flavorjones explained is exactly what I was saying.

raybradley:
Ok think I got it...omitted `gem update bundler` from Dockerfile. Thanks for the assist!

flavorjones:
OK, new version of rubygems and bundler dropped today but there's still problems on musl platforms, I've filed a bug upstream: https://github.com/rubygems/rubygems/issues/5914

flavorjones:
Clarifying my last comment: although there is still an issue with `gem install` in rubygems 3.3.22, the corresponding version of bundler, 2.3.22, does work correctly and installs the `x86_64-linux` platform gem on musl systems.

So I don't think we need this PR or a `x86_64-linux-musl` platform gem. I'm going to close this without merging.

PR #199, opened 2022-09-03
feat: support polling for watching files

flavorjones:
Polling was added upstream in 3.1.0 to solve timing issues with systems that don't fully support file system events (see https://github.com/tailwindlabs/tailwindcss/pull/7725).

This builds on the commits at #198 to support a `poll` argument to the `tailwindcss:watch` rake task:

```
rails tailwindcss:watch[poll]
```

Because we're using rake task arguments, we can combine `poll` with the `debug` argument from #198 like so:

```
rails tailwindcss:watch[debug,poll]
```

This PR supersedes #168

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
 
 ## Unreleased
 
-* Correctly handle paths with embedded spaces. [#184](https://github.com/rails/tailwindcss-rails/issues/184)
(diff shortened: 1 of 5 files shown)


dhh (approved):

dhh:
TIL about using rake task arguments like this! 👍

PR #226, opened 2022-11-14
feat: fall back to locally-installed tailwindcss executables

### LINKED ISSUE #224: Pointing to the tailwindcss executable
I'm running NixOs and due to multiple reasons I cannot install platform-native gems. However, I _can_ install the tailwindcss executable on its own. However, this gem does not try checking PATH if the included version does not exist.

I understand that this is probably for easier maintenance and is a good idea. However, could a configuration/escape hatch be added that would allow me to point it at the tailwindcss executable? The onus would be on me to make sure the versions match up etc...

This would make it possible to use on NixOS while still keeping the niceties of the gem.

flavorjones:
If a user has installed the `ruby` platform gem, but has the `tailwindcss` executable installed in their path, then we should fall back to that.

Closes #224

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 ## next / unreleased
 
 * Update to [Tailwind CSS v3.3.0](https://github.com/tailwindlabs/tailwindcss/releases/tag/v3.3.0) by @tysongach
+* Users can use a locally-installed `tailwindcss` executable by setting a `TAILWINDCSS_INSTALL_DIR` environment variable. (#224, #226) by @flavorjones
 
 
 ## v2.0.25 / 2023-03-14
--- a/README.md
+++ b/README.md
@@ -9,7 +9,34 @@ With Rails 7 you can generate a new application preconfigured with Tailwind by u
 1. Run `./bin/bundle add tailwindcss-rails`
 2. Run `./bin/rails tailwindcss:install`
 
-This gem wraps [the standalone executable version](https://tailwindcss.com/blog/standalone-cli) of the Tailwind CSS v3 framework. These executables are platform specific, so there are actually separate underlying gems per platform, but the correct gem will automatically be picked for your platform. Supported platforms are Linux x64, macOS arm64, macOS x64, and Windows x64. (Note that due to this setup, you must install the actual gems – you can't pin your gem to the github repo.)
+This gem wraps [the standalone executable version](https://tailwindcss.com/blog/standalone-cli) of the Tailwind CSS v3 framework. These executables are platform specific, so there are actually separate underlying gems per platform, but the correct gem will automatically be picked for your platform.
+
+Supported platforms are:
+
+- arm64-darwin (macos-arm64)
+- x64-mingw32 (windows-x64)
+- x64-mingw-ucr (windows-x64)
+- x86_64-darwin (macos-x64)
+- x86_64-linux (linux-x64)
+- aarch64-linux (linux-arm64)
+- arm-linux (linux-armv7)
+
+
+### Using a local installation of `tailwindcss`
+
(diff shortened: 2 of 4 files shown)


flavorjones:
@TheNeikos do you want to give this branch a try and let me know if it works for you?

Here's how you can try it out:

- checkout this branch (`flavorjones-fallback-to-manually-installed`)
- `bundle install && bundle exec rake gem:ruby`
- `gem install --local pkg/tailwindcss-rails-2.0.21.gem`

TheNeikos:
Sadly it still does not work :c 

This is because you are explicitly checking for wrapper scripts and ignore them? I don't think that is necessary, since users can decide themselves how they want their executables to be called.

In my case for example, and quite usual for NixOS, `tailwindcss` is a wrapper script that sets a bunch of env variables before calling the actual binary.

flavorjones:
Can you help me understand what "does not work" looks like for you? Can you give me more information about _how_ it doesn't work?

I chose to ignore wrapper scripts because Rubygems will install a `tailwindcss` wrapper script that's in the user's path, and so I need some mechanism to avoid infinite recursion (calling the `tailwindcss` rubygems shim, which calls this gem's `tailwindcss` script, which then call the rubygems shim again). If you have any ideas on how to do this in a robust manner that works on your system, please let me know.

flavorjones:
Maybe I'm overthinking this, would a suitable alternative be to set an environment variable to the path of your "real" tailwindcss installation?

Something like this in your .envrc for example:

```
TAILWINDCSS_RAILS_LOCAL_PATH=/path/to/nix/installation/bin/tailwindcss
```

Would that solve your use case?

TheNeikos:
> Can you help me understand what "does not work" looks like for you? Can you give me more information about _how_ it doesn't work?
> 
> I chose to ignore wrapper scripts because Rubygems will install a `tailwindcss` wrapper script that's in the user's path, and so I need some mechanism to avoid infinite recursion (calling the `tailwindcss` rubygems shim, which calls this gem's `tailwindcss` script, which then call the rubygems shim again). If you have any ideas on how to do this in a robust manner that works on your system, please let me know.

Sorry I didn't go into more detail. But basically, same behaviour as before. It just tells me that I need to have it installed (since it _is_ just a wrapper script I have in my PATH).

My first question would be, why does the rubygems shim have the same name? Could it be removed/renamed?

> Would that solve your use case?

That would 100% solve it! I can easily set things like that.

flavorjones:
> why does the rubygems shim have the same name

For the same reason that Nix's shim has the same name?

> That would 100% solve it

OK, I'll overhaul this PR with that approach.

TheNeikos:
> For the same reason that Nix's shim has the same name?

Haha, fair enough! I just wouldn't expect a rubygem to install an executable with the same name as the one it tries to catch only to say that it is not installed!

In the NixOS case its to make sure tailwind can find all the plugins it needs and is more of a redirect to the actual executable.

piperswe:
Another Nix user here: this is the main blocker for me moving my app's build process entirely under Nix, is there anything I can do to help move it along?

piperswe:
I just ended up creating a new PR (#259) with the environment variable override approach

nanaya:
fwiw, it's also a problem for people running on "unsupported" platform (freebsd etc). I just override the whole function for now. Actual tailwindcss can be installed no problem on platforms with node using npm/yarn.

```ruby
# config/initializers/00tailwind_freebsd.rb
module Tailwindcss
  module Commands
    def self.executable(*)
      File.expand_path(File.join(__dir__, "..", "..", "node_modules", ".bin", "tailwindcss"))
    end
  end
end
```

flavorjones:
I've updated this branch with an environment-variable-setting approach.

PR #229, opened 2022-12-03
Fix generating nested models

### LINKED ISSUE #227: View files generated by scaffold cause `ActionView::MissingTemplate` when the Model name is nested
View files generated by scaffold cause `ActionView::MissingTemplate` when the Model name is nested.

```
$ rails new myapp --css=tailwind
$ cd myapp
$ bin/rails generate scaffold Foo/Bar title:string body:text
$ bin/rails db:migrate
$ bin/rails r 'Foo::Bar.create'
$ bin/rails s
$ curl -is http://localhost:3000/foo/bars | grep MissingTemplate
    ActionView::MissingTemplate in
</div><div class="line active">        raise(MissingTemplate.new(self, path, prefixes, partial, details, details_key, locals))
```

Probably [this line](https://github.com/rails/tailwindcss-rails/blob/main/lib/generators/tailwindcss/scaffold/scaffold_generator.rb#L25) is wrong, and `singular_name` should be used instead of `singular_table_name` like [rails/rails](https://github.com/rails/rails/blob/ade4e7a907e4fd36f0835f9c4150817fb978dfbd/railties/lib/rails/generators/erb/scaffold/scaffold_generator.rb#L25), but I'm not 100% sure.

dixpac:
This PR fixes error when generating nested models such as:

`bin/rails g scaffold admin/role name description`

Closes #227

### DIFF
--- a/lib/generators/tailwindcss/scaffold/scaffold_generator.rb
+++ b/lib/generators/tailwindcss/scaffold/scaffold_generator.rb
@@ -22,7 +22,7 @@ module Tailwindcss
           end
         end
 
-        template "partial.html.erb", File.join("app/views", controller_file_path, "_#{singular_table_name}.html.erb")
+        template "partial.html.erb", File.join("app/views", controller_file_path, "_#{singular_name}.html.erb")
       end
 
       private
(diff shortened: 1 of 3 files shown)


flavorjones:
@dhh please review this when you get a moment!

PR #230, opened 2022-12-05
Clarify how to run Tailwind command in CI

ghiculescu:
The current docs are a bit confusing, because only `rake` tasks get enhanced by this library, but there's also `rails` tasks that look the same.

I tried to make it a bit more clear.

Disclaimer: it's possible I've gotten confused and there's a much simpler recommended approach... I'd love for that to be true!

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ The `tailwindcss:build` is automatically attached to `assets:precompile`, so bef
 
 ### Building for testing
 
-The `tailwindcss:build` task is automatically attached to the `test:prepare` task. The `test:prepare` task is run before some test tasks (e.g. `test:all` and `test:controllers`), but not before the bare `test` task.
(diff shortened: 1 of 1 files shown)


flavorjones:
To simplify the message, should we just recommend one way to do it, and that's `bin/rails test:prepare test`?

ghiculescu:
Yes, that's much better.

ghiculescu:
Just one suggestion. I hope that https://github.com/rails/rails/pull/46664 will make all this irrelevant! Thanks for your help @flavorjones

flavorjones:
Thanks for diving into the Rails tasks, I had just started down that path. homer-disappearing-into-a-bush.gif

ghiculescu:
> Thanks for diving into the Rails tasks, I had just started down that path. homer-disappearing-into-a-bush.gif

Well I haven't quite got it working right, see my comment on the PR. Help is welcome!! 🙏

Issue #127, opened 2022-01-06
Dynamic classes generated in erb not recognized


GuidoCaldara:
Hi everyone. I have a Rails 7 app with tailwind where I'm doing something like this:

```
  @list = [{name: "Some", width: "3/12"}, {name: "other", width: "6/12"}]
  
  # In the view
  <%= render 'path/to/mypartial': list: @list %>
  
  # In the partial
  <% list.each do |item|%>
    <div class="w-<%= item[:width] %>">
      <%= item[:name] %>
    </div>
  <% end %>
```
The view generates the proper HTML (ex: `<div class="w-6/12">`), but the classes are not recognized in the browser. 

I've realized after I opened the issue that to make it work I had to add all the dynamic classes to a `purge/safelist` to the `tailwind.config.js`.
I'm new to Tailwind so I'm not sure if there's any workaround to this. Thanks :)

dhh:
This is a standard TW constraint. Documented it here: https://github.com/rails/tailwindcss-rails/blob/main/README.md#class-names-must-be-spelled-out

elalemanyo:
Hi,
Since tailwind requires classes to be spelled out we lose dynamics in the views.
In case that the number of possible classes are not too many maybe a good hack would be to add a unused view (`_trigger_dynamic_ generated_classes.html.erb`) where you can write all `dynamic` classes.
I know it is already closed, but in case it might help someone else...
Thanks

fwuensche:
If you can figure an exhaustive list of the dynamic classes you're generating, you can also whitelist them in your `tailwind.config.js` file. To illustrate, let's say you have an erb file containing: 

```rb
link_to 'Show', project_path, class: "text-#{color}-700"
```

You'd just need to list your dynamically generated values, such as:

```js
module.exports = {
  purge: {
    safelist: [
      'text-green-700',
      'text-red-700',
    ],
  },
}
```

elalemanyo:
Thank you @fwuensche, did not know about this functionality in tailwind. I have created a [pull request](https://github.com/rails/tailwindcss-rails/pull/233) adding your explanation.
I have to admit that adding a unused view allows you to use the full power of `.erb`, but your solution is more academic 🙂

PR #233, opened 2022-12-12
Add explanation for programmatically composed class names to use the safelist option

elalemanyo:
I think it is interesting to have this documented. It's something that has already come up in an [issue](https://github.com/rails/tailwindcss-rails/issues/127), but it's better not to hide it so much.


flavorjones:
Thank you for your contribution!

PR #239, opened 2023-01-09
Simplify "Building for testing" [merge after Rails 7.1]

ghiculescu:
Now that https://github.com/rails/rails/pull/46664 is merged, we can simplify the testing instructions.

Technically these instructions aren't true until Rails 7.1, so I'm happy to leave the PR open until then if that's preferred.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,13 +28,8 @@ The `tailwindcss:build` is automatically attached to `assets:precompile`, so bef
 
 ### Building for testing
 
-The `tailwindcss:build` task is automatically attached to the `test:prepare` Rake task. The `test:prepare` task is run before some test tasks (e.g. `test:all` and `test:controllers`), but not before the bare `test` task.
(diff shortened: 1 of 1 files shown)


flavorjones:
:partying_face: Congrats on making that change upstream!

I still feel like people are slow enough to upgrade that it may make sense to keep _some_ version of this explanation around for a while. WDYT?

ghiculescu:
Yes I agree, I mostly didn’t want to forget about it. We can just keep this PR open for a while. Or are you thinking we have “if you use Rails 7.1, do X, otherwise do Y” style instructions?

flavorjones:
I meant, let's leave this unmerged until Rails 7.1 ships.

ghiculescu:
@flavorjones is this okay to merge now?

flavorjones:
@ghiculescu Thank you for the reminder! Merging.

Issue #164, opened 2022-04-20
Overriding scaffold erb templates


craic:
Rails 7.0.2.3 ruby 3.0.0p0 ubuntu

I want to modify the erb templates used in scaffolding but the tailwindcss template engine doesn't pick up custom templates from lib/templates/erb/scaffold

Does it use a different location ? I don't see an option in 'rails g scaffold --help' where I can specify that explicitly

Thanks !

dhh:
You can configure which files are watched in tailwind.config.js.

mabras:
> You can configure which files are watched in tailwind.config.js.

Please, can you double-check your answer? It sounds to me like it has nothing to do with the issue.

We are trying to override the erb templates generated by `rails g scaffold [resource]`.
Usually, we put the erb files in locations like:
```
lib/templates/erb/scaffold/*.html.erb
```
Also, I have tried this path:
```
lib/generators/tailwindcss/scaffold/templates/*.html.erb
```
Both are not working.

Thanks @dhh

dhh:
Ah I read it as a watching question. This gem is already overriding the scaffold templates, so that's probably why. Would be happy to look at a PR if you find a way to make this compatible with overwritten templates.

craic:
This source_root line here is overriding the default behaviour

https://github.com/rails/tailwindcss-rails/blob/47f636b12880dc07c9c6210a7c71658186844c5f/lib/generators/tailwindcss/scaffold/scaffold_generator.rb#L1-L9

If I comment that out then I can pick up from 
`lib/templates/erb/scaffold/*.html.erb`

Maybe an if / else clause that looks to see if the app has defined custom templates then use those, if not then use the ones in tailwindcss

Unfortunately I don't see how where the custom template path is defined in the regular generator.
If you can give me a pointer then I will try it out

luiskhernandez:
@craic Not sure if this could be merged into the gem, but overriding the class method `source_root` did the trick.

I added this file `lib/generators/tailwindcss/scaffold/scaffold_generator.rb`

```
require 'rails/generators/erb/scaffold/scaffold_generator'
require "rails/generators/resource_helpers"

module Tailwindcss
  module Generators
    class ScaffoldGenerator < Erb::Generators::ScaffoldGenerator

      def self.source_root
        default_path = Rails.root.join("lib", "templates", "erb", "scaffold")
        return default_path if Dir.exists?(default_path)
        temp = File.expand_path("../templates", __FILE__)
      end
    end
[...]

PR #247, opened 2023-02-11
Prioritize custom view templates

mojodojo101:
Current behavior:
Custom templates are ignored by the tailwindcss-rails generator

Expected behavior:
View templates under "#{RAILS_ROOT}/lib/templates/erb/scaffold" should be prioritized by the generator.

https://github.com/rails/tailwindcss-rails/issues/164

### DIFF
--- a/lib/generators/tailwindcss/scaffold/scaffold_generator.rb
+++ b/lib/generators/tailwindcss/scaffold/scaffold_generator.rb
@@ -6,7 +6,16 @@ module Tailwindcss
     class ScaffoldGenerator < Erb::Generators::ScaffoldGenerator
       include Rails::Generators::ResourceHelpers
 
-      source_root File.expand_path("../templates", __FILE__)
+
+      def self.source_root
+        fallback_path=File.expand_path("../templates", __FILE__)
+        if Rails.root.nil?
+          return fallback_path
+        else
+          custom_scaffold_templates_path = File.join(Rails.root,"lib", "templates", "erb", "scaffold")
+          return Dir.exist?(custom_scaffold_templates_path)? custom_scaffold_templates_path : fallback_path
+        end
+      end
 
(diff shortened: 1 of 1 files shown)


flavorjones:
@mojodojo101 Thanks for submitting this. CI is failing, can you look into it, please?

mojodojo101:
> @mojodojo101 Thanks for submitting this. CI is failing, can you look into it, please?

Hmm, i guess Rails.root isnt initialized in those tests, would it be fine to simply check if its nil with an if statement or is that piece of code i wrote just rubbish?

flavorjones:
Can you please remove the version bump and changelog entry from this PR to avoid conflicts?

Would it be possible to write a test to describe the behavior you're introducing?

mojodojo101:
> 

I removed the Changelog and Version bump 😄 . I dont know much about writing tests/ integration tests so i might have to look into that once i have some time on my hands. 

The tldr of the behavior introduced here is:
- Check if there is a /lib/templates/erb/scaffold  folder in rails
- If that is the case use the templates in that scaffold folder rather than the folder provided by tailwindcss-rails 


Following the procedure that is described here

https://guides.rubyonrails.org/generators.html :
... Another common use of custom templates is overriding the [default scaffold view templates](https://github.com/rails/rails/tree/main/railties/lib/rails/generators/erb/scaffold/templates). You can override any of these by creating the appropriate file (e.g. index.html.erb, show.html.erb, etc) in lib/templates/erb/scaffold.

---
At the moment it just checks if the folder is there, if we want file granularity, this commit wont do just fyi.

rgarver:
I just hit this issue. Is there any additional work required to make this mergeable?

flavorjones:
I'm closing this in preference of #314 which is a simpler fix and has test coverage.

PR #248, opened 2023-02-12
allow users to always generate unminified CSS

### LINKED ISSUE #244: [feature] allow users to always generate unminified CSS
### Discussed in https://github.com/rails/tailwindcss-rails/discussions/243

<div type='discussions-op-text'>

<sup>Originally posted by **the-pulli** February  2, 2023</sup>
Are there any strategies how to use Tailwind with yui as a CSS compressor?

If I enable yui, the already minified tailwind.css from (app/assets/builds) gets again minified (and messed up) by yui and copied as an asset to public/assets.

Can I configure Rails to exclude the tailwind.css from being minified again?

Any help is appreciated, thanks 🙃</div>

### Actions

Because `assets:precompile` doesn't allow injection of the `debug` flag, this is currently hard for users to _always_ do. Let's figure out how to enable users to always generate unminified tailwind assets.

flavorjones:
Discussed in #243.
Closes #244.

This PR allows users to set an environment variable `TAILWINDCSS_RAILS_NO_MINIFY` to turn off minification.

### DIFF
--- a/lib/tailwindcss/commands.rb
+++ b/lib/tailwindcss/commands.rb
@@ -74,11 +74,20 @@ module Tailwindcss
       end
 
       def debug_option(rake_task_args_extras)
-        rake_task_args_extras.include?("debug")
+        rake_task_args_extras.include?("debug") || truthy_string_value(ENV['TAILWINDCSS_RAILS_NO_MINIFY'])
       end
 
       def poll_option(rake_task_args_extras)
         rake_task_args_extras.include?("poll")
+      end
+
+      def truthy_string_value(value)
+        if !value.present? ||
+           ["f", "F", "false", "FALSE", "off", "OFF", "0"].include?(value)
+          false
+        else
+          true
+        end
       end
     end
   end
--- a/lib/tailwindcss/commands.rb
+++ b/lib/tailwindcss/commands.rb
@@ -72,6 +72,14 @@ module Tailwindcss
           command << "-p" if poll
         end
       end
+
+      def debug_option(rake_task_args_extras)
+        rake_task_args_extras.include?("debug")
+      end
+
+      def poll_option(rake_task_args_extras)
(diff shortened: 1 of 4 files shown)


lib/tailwindcss/commands.rb
```diff
@@ -72,6 +72,23 @@ def watch_command(poll: false, **kwargs)
...
+      def debug_option(rake_task_args_extras)
+        rake_task_args_extras.include?("debug") || truthy_string_value(ENV['TAILWINDCSS_RAILS_NO_MINIFY'])
+      end
+
+      def poll_option(rake_task_args_extras)
+        rake_task_args_extras.include?("poll")
+      end
+
+      def truthy_string_value(value)
+        if !value.present? ||
```
rafaelfranca, line 85: Do we have `present?` available? The core_ext is not being required in this gem. I think it only works in tests because the app is configured to require all core_ext.

dhh:
I don't see the point? I see the request, but not the purpose.

flavorjones:
@the-pulli Can you explain your use case a bit more, please?

the-pulli:
@flavorjones Sure thing. I have a project which uses TW and custom CSS. If I don't use yui as CSS compressor only my TW CSS gets minified and my custom CSS not. If I use yui  as compressor TW gets again minified through yui and it get messed up. My custom CSS gets minified normally in this procedure. I was looking for a way to minify both TW and my custom CSS for production.

dhh:
I think this should be fixed elsewhere rather than in TW. The minifier should be able to opt out of being applied to a given CSS. Rather than TW knowing about not having to minify.

the-pulli:
@dhh So if I understood you right, there is currently no way to do this in Rails? It has to be implemented (preferable the code handling compression of the files), but not in TW?

PR #253, opened 2023-03-02
feat: tailwind will not minify when Rails CSS compression is on

flavorjones:
Looking for feedback on this approach.

This will avoid double-compression which breaks styles.

Fixes #243

See https://github.com/rails/rails/issues/47553


dhh:
I do like this. Very unobtrusive. Good from me 👍

dhh (approved):

flavorjones:
Shipped in v2.0.24.

PR #254, opened 2023-03-03
Added 2 official Tailwind plugins

Kentasmic:
Tailwind has 5 official plugins but only the first 3 are installed by default. 

This pull request adds the final 2 plugins (line clamp and container queries) to the installer for config/tailwind.config.js.


flavorjones:
Shipped in v2.0.25. Thank you!

PR #255, opened 2023-03-04
Fixes #252 windows support for dev command and foreman proc file

### LINKED ISSUE #252: Need help to fix tailwindcss-rails support on Windows
Hi,
I try to support this gem on Windows plateform.
I have detected 2 problems: 

1. bin/dev is a linux shell command
I suggest to add a new file named **bin/dev.bat**
```
@echo off

set gem_installed=

for /f "delims=" %%a in ('gem list foreman -i') do @set gem_installed=%%a

if %gem_installed%==false (
  echo "Installing foreman..."
  gem install foreman
)

foreman start -f Procfile.dev %*
```
2. command **bin/rails** in **Procfile.dev** is not recognized
Replace **bin/rails** by **ruby bin/rails**
I suggest to create a file named Procfile.dev.windows and copy this file only if windows platform.
Problem: I don't have found where the copy of files are done.
Can somebody help me ?
After this I can submit a pull request.
Best regards.

espinosa:
New file **bin/dev.bat**
Create procfile.dev with **ruby bin/rails** instead of **rails/bin**

### DIFF
--- a/lib/install/dev.bat
+++ b/lib/install/dev.bat
@@ -1,0 +1,12 @@
+@echo off
+
+set gem_installed=
+
+for /f "delims=" %%a in ('gem list foreman -i') do @set gem_installed=%%a
+
+if %gem_installed%==false (
+  echo "Installing foreman..."
+  gem install foreman
+)
+
+foreman start -f Procfile.dev %*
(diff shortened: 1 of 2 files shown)


dhh:
I don't see us supporting and maintaining bat files. We don't do that in any other gem. I'd suggest using WSL2 or similar to be able to use the core unix versions.

PR #259, opened 2023-03-30
feat: allow overriding executable with environment variable

### LINKED ISSUE #224: Pointing to the tailwindcss executable

piperswe:
This allows users to provide a path to their own `tailwindcss` executable, which can be handy in cases where the pre-built ones won't work (e.g. in bundix-based Nix setups).

Closes #224
Replaces/closes #226

### DIFF
--- a/lib/tailwindcss/commands.rb
+++ b/lib/tailwindcss/commands.rb
@@ -25,11 +25,17 @@ module Tailwindcss
           MESSAGE
         end
 
-        exe_path = Dir.glob(File.expand_path(File.join(exe_path, "*", "tailwindcss"))).find do |f|
-          Gem::Platform.match(Gem::Platform.new(File.basename(File.dirname(f))))
+        # Let the user override the executable path with an environment variable
+        # Useful for when running under Nix
+        exe_file = ENV["TAILWINDCSS_RAILS_EXECUTABLE_PATH"]
+
+        if exe_file.nil? || exe_file.empty?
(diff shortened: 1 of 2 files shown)


flavorjones:
Hi, @piperswe, thanks for the PR. I'm not sure the testing coverage here is thorough enough -- see #226 for some examples of the kinds of tests that I think are necessary when we're trying to ensure the native binaries will work correctly.

I may end up just updating #226 with the env var approach as I mentioned there ... though if you want to take a stab at increasing the test coverage here I'd be happy to give more explicit feedback.

Either way, I appreciate the nudge, and I will try to ship something this weekend.

nanaya:
Platform check needs to be skipped when env is specified otherwise it doesn't help people running on unsupported platforms but do have nodejs which can run tailwindcss.

flavorjones:
@piperswe Thanks again for the PR and for nudging me on this. I've updated #226 with a similar approach to what you've done here, but with a bit more test coverage for a few different edge cases. I'm going to close this in preference to that PR.

PR #262, opened 2023-04-20
Handle `tailwindcss:watch[always]`

GoodForOneFare:
Passes `-w always` along to tailwindcss, which keeps the watch going even if stdin closes (https://github.com/tailwindlabs/tailwindcss/pull/9966).

### DIFF
--- a/lib/tailwindcss/commands.rb
+++ b/lib/tailwindcss/commands.rb
@@ -83,9 +83,10 @@ module Tailwindcss
         end
       end
 
-      def watch_command(poll: false, **kwargs)
+      def watch_command(always: false, poll: false, **kwargs)
         compile_command(**kwargs).tap do |command|
           command << "-w"
+          command << "always" if always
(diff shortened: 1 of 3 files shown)


rafaelfranca:
Any reason why this can't be the default of `tailwindcss:watch`?

GoodForOneFare:
@rafaelfranca the default behaviour is IMO better.  In most environments, the default mode cleans up its own orphaned processes.  The tailwind developers added `--watch always` with reluctance, and I think it's only useful for niche environments where developers don't have control of their Docker config.

flavorjones:
@GoodForOneFare I could use your help to understand what we should add to the "troubleshooting" section to help users understand when to use this feature. What will people see if they need to set this?

flavorjones:
See https://github.com/rails/tailwindcss-rails/commit/e665b097ae51e4e4d388370352a6390d41e685fd

PR #271, opened 2023-06-20
Enhance spec:prepare command with tailwindcss:build command

rmehner:
jsbundling-rails (https://github.com/rails/jsbundling-rails/pull/128) and cssbundling-rails (https://github.com/rails/cssbundling-rails/pull/116) do the same and it's useful for people who are using RSpec.


flavorjones (approved):

flavorjones:
I continue to suspect that there's an opportunity to put common behavior across the asset-management gems into shared code.

rmehner:
> I continue to suspect that there's an opportunity to put common behavior across the asset-management gems into shared code.

Yup, definitely. We've also been bitten by the subtle difference between tailwindcss-rails and cssbundling-rails with Tailwind a couple of times.

PR #280, opened 2023-10-09
Fix deprecation warnings in 3.5.0

lylo:
Running Ruby 3.3.0-preview2 and Ruby Gems 3.5.0.dev I am seeing the following deprecation warning:

```
NOTE: Gem::Platform.match is deprecated; use Gem::Platform.match_spec? or match_gem? instead. It will be removed in Rubygems 4
Gem::Platform.match called from /Users/[me]/dev/tailwindcss-rails/lib/tailwindcss/commands.rb:37.
```

This patch replaces calls to `Gem::Platform#match` to `Gem::Platform#match_gem?` as recommended in [the source](https://github.com/rubygems/rubygems/blob/28f813ba97d6ed3a8ede427b0dc5e7fab47aa02f/lib/rubygems/platform.rb#L27).

This update passes `nil` as the second argument to `match_gem?`. The implementation of `match_gem?` appears to ignore the second argument (`gem_name`) and it's actually identical to the deprecated `match` method:

```ruby
  def self.match(platform)
    match_platforms?(platform, Gem.platforms)
  end

  def self.match_gem?(platform, gem_name)
    # NOTE: this method might be redefined by Ruby implementations to
    # customize behavior per RUBY_ENGINE, gem_name or other criteria.
    match_platforms?(platform, Gem.platforms)
  end
```

### DIFF
--- a/lib/tailwindcss/commands.rb
+++ b/lib/tailwindcss/commands.rb
@@ -34,7 +34,7 @@ module Tailwindcss
             MESSAGE
           end
         else
-          if Tailwindcss::Upstream::NATIVE_PLATFORMS.keys.none? { |p| Gem::Platform.match(Gem::Platform.new(p)) }
+          if Tailwindcss::Upstream::NATIVE_PLATFORMS.keys.none? { |p| Gem::Platform.match_gem?(Gem::Platform.new(p), nil) }
             raise UnsupportedPlatformException, <<~MESSAGE
               tailwindcss-rails does not support the #{platform} platform
               Please install tailwindcss following instructions at https://tailwindcss.com/docs/installation
@@ -42,7 +42,7 @@ module Tailwindcss
           end
 
           exe_file = Dir.glob(File.expand_path(File.join(exe_path, "*", "tailwindcss"))).find do |f|
-            Gem::Platform.match(Gem::Platform.new(File.basename(File.dirname(f))))
+            Gem::Platform.match_gem?(Gem::Platform.new(File.basename(File.dirname(f))), nil)
           end
         end
 
--- a/lib/tailwindcss/commands.rb
+++ b/lib/tailwindcss/commands.rb
@@ -3,6 +3,7 @@ require_relative "upstream"
 module Tailwindcss
   module Commands
     DEFAULT_DIR = File.expand_path(File.join(__dir__, "..", "..", "exe"))
+    GEM_NAME = "tailwindcss-rails"
 
     # raised when the host platform is not supported by upstream tailwindcss's binary releases
     class UnsupportedPlatformException < StandardError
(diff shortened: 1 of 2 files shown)


flavorjones:
@lylo Ah, thank you so much for the PR! I hadn't seen these warnings yet. I'll take a look at the PR later today.

lib/tailwindcss/commands.rb
```diff
@@ -34,15 +34,15 @@ def executable(exe_path: DEFAULT_DIR)
             MESSAGE
           end
         else
-          if Tailwindcss::Upstream::NATIVE_PLATFORMS.keys.none? { |p| Gem::Platform.match(Gem::Platform.new(p)) }
+          if Tailwindcss::Upstream::NATIVE_PLATFORMS.keys.none? { |p| Gem::Platform.match_gem?(Gem::Platform.new(p), nil) }
```
flavorjones, line 37: The second argument here should be the gem name, `"tailwindcss-rails"`.

Although this argument isn't used in the CRuby implementation, other implementations (specifically TruffleRuby) reserve the right to re-implement this method with special cases for specific gems. If you'd like to learn more about the history behind this curious design decision, it's all at https://github.com/rubygems/rubygems/pull/3817

lib/tailwindcss/commands.rb
```diff
@@ -34,15 +34,15 @@ def executable(exe_path: DEFAULT_DIR)
...
             raise UnsupportedPlatformException, <<~MESSAGE
               tailwindcss-rails does not support the #{platform} platform
               Please install tailwindcss following instructions at https://tailwindcss.com/docs/installation
             MESSAGE
           end
 
           exe_file = Dir.glob(File.expand_path(File.join(exe_path, "*", "tailwindcss"))).find do |f|
-            Gem::Platform.match(Gem::Platform.new(File.basename(File.dirname(f))))
+            Gem::Platform.match_gem?(Gem::Platform.new(File.basename(File.dirname(f))), nil)
```
flavorjones, line 45: See previous comment.

flavorjones (changes requested):

flavorjones:
The user journey tests are failing for unrelated reasons ... I'm going to fix on `main` and will ask you to rebase when it's ready.

flavorjones:
@lylo Because I'd like to get a release out today with this and #281, I went ahead and made the small changes I requested in my review.

lylo:
> @lylo Because I'd like to get a release out today with this and #281, I went ahead and made the small changes I requested in my review.

No worries! Thank you for explaining the gem name param too 🙏

flavorjones:
https://github.com/rails/tailwindcss-rails/releases/tag/v2.0.31 is out! Thank you again!

PR #281, opened 2023-10-10
Run the server with remote debugging enabled

duduribeiro:
This commit adds the `--open` option so a remote session can be started using `rdbg --attach`. This allows to debug inside a foreman process.

This will replicate what we have on jsbundling-rails https://github.com/rails/jsbundling-rails/blob/main/lib/install/Procfile.dev#L1 and on cssbundling-rails https://github.com/rails/cssbundling-rails/blob/main/lib/install/Procfile_for_bun.dev#L1

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## next / unreleased
+
+* Update Procfile.dev to run foreman with `--open` option allowing remote sessions with `rdbg --attach` by @duduribeiro
 
 ## v2.0.30 / 2023-07-13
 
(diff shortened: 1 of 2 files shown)


flavorjones:
Quick note that the user journey tests are failing for unrelated reasons (will be fixed on main when https://github.com/rails/tailwindcss-rails/pull/282 is merged).

flavorjones:
@duduribeiro Thank you for opening this! Looks great, works fine for me.

flavorjones:
https://github.com/rails/tailwindcss-rails/releases/tag/v2.0.31 is out! Thank you again!

PR #289, opened 2023-10-27
dep: rubygems >= 3.2.0 for Platform#match_gem?

### LINKED ISSUE #286: v2.0.31 Fails on ruby 2.7.x
PR https://github.com/rails/tailwindcss-rails/pull/280 replaces ```Gem::Platform#match``` to ```Gem::Platform#match_gem?``` This fixes deprecation warnings but breaks on ruby 2.7.? because match_gem? does not exist there (at least not on 2.7.3).

This causes errors like these:

```
bin/rails tailwindcss:watch
rails aborted!
NoMethodError: undefined method `match_gem?' for Gem::Platform:Class

Tasks: TOP => tailwindcss:watch
```

and

when running ```bin/rails s```
```
ActionView::Template::Error (The asset "tailwind.css" is not present in the asset pipeline.
):
     5:     <meta name="viewport" content="width=device-width,initial-scale=1">
     6:     <%= csrf_meta_tags %>
     7:     <%= csp_meta_tag %>
     8:     <%= stylesheet_link_tag "tailwind", "inter-font", "data-turbo-track": "reload" %>
     9: 
    10:     <%= stylesheet_link_tag "application", "data-turbo-track": "reload" %>
    11:     <%= javascript_importmap_tags %>
  
app/views/layouts/application.html.erb:8
```

## Work around

Update your gemfile to fix tailwindcss-rails to version 2.0.30 and run bundle update.

flavorjones:
Closes #286

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,6 +26,7 @@ jobs:
       - uses: ruby/setup-ruby@v1
         with:
           ruby-version: ${{matrix.ruby}}
+          rubygems: latest
           bundler: latest
           bundler-cache: true
       - name: Run tests
--- a/tailwindcss-rails.gemspec
+++ b/tailwindcss-rails.gemspec
@@ -14,6 +14,8 @@ Gem::Specification.new do |spec|
     "rubygems_mfa_required" => "true"
   }
 
+  spec.required_rubygems_version = ">= 3.2.0" # for Gem::Platform#match_gem?
+
   spec.files = Dir["{app,lib}/**/*", "MIT-LICENSE", "Rakefile", "README.md"]
   spec.bindir = "exe"
   spec.executables << "tailwindcss"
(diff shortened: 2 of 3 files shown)


flavorjones:
@henrikbjorn That might have been a good idea, but I can't go back in time and fix it now. I'd suggest pinning your app to 2.0.30  if you can't upgrade rubygems.

Out of curiosity, can you share more about your situation and why you're not able to upgrade? Rubygems 3.2.0 is over three years old at this point I think should be shipped in all versions of Ruby that are not EOL.

henrikbjorn:
> @henrikbjorn That might have been a good idea, but I can't go back in time and fix it now. I'd suggest pinning your app to 2.0.30 if you can't upgrade rubygems.
> 
> Out of curiosity, can you share more about your situation and why you're not able to upgrade? Rubygems 3.2.0 is over three years old at this point I think should be shipped in all versions of Ruby that are not EOL.

Can't be specific.

But I think some might have old infrastructures, or stuck with gems that haven't been updated in a while etc. 

How about doing a point release for 2.0.x that checks for 'respond_to?` or something?

flavorjones:
There are two known workarounds:

- upgrade rubygems
- pin this gem to 2.0.30

Because there are ways around this issue, and because this issue affects so few people, writing and testing a third workaround and making a release feels like a relatively low priority for me, at least without additional information.

PR #296, opened 2023-12-02
ci: more thorough user journey, integrated with rails main

flavorjones:
- make `test/integration/user_journey_test.sh` a bit more thorough
- add the user journey to the "upstream" pipeline, and test against rails main
- update CI generally to ruby 3.2 or 3.3 (head)
- fix the musl native installation test; the container needs build-base to compile other gems


flavorjones:
For context, this PR was partly in response to https://github.com/rails/rails/pull/49679

PR #300, opened 2023-12-21
Add puma plugin

npezza93:
This adds a puma plugin that runs watch mode only in development.

### DIFF
--- a/lib/puma/plugin/tailwindcss.rb
+++ b/lib/puma/plugin/tailwindcss.rb
@@ -1,0 +1,63 @@
+require "puma/plugin"
+
+Puma::Plugin.create do
+  attr_reader :puma_pid, :tailwind_pid, :log_writer
+
+  def start(launcher)
+    @log_writer = launcher.log_writer
+    @puma_pid = $$
+    @tailwind_pid = fork do
+      Thread.new { monitor_puma }
+      system(*Tailwindcss::Commands.watch_command)
+    end
+
+    launcher.events.on_stopped { stop_tailwind }
+
+    in_background do
+      monitor_tailwind
+    end
+  end
+
+  private
+    def stop_tailwind
+      Process.waitpid(tailwind_pid, Process::WNOHANG)
+      log "Stopping tailwind..."
+      Process.kill(:INT, tailwind_pid) if tailwind_pid
+      Process.wait(tailwind_pid)
+    rescue Errno::ECHILD, Errno::ESRCH
+    end
+
+    def monitor_puma
+      monitor(:puma_dead?, "Detected Puma has gone away, stopping tailwind...")
+    end
+
+    def monitor_tailwind
+      monitor(:tailwind_dead?, "Detected tailwind has gone away, stopping Puma...")
+    end
+
+    def monitor(process_dead, message)
+      loop do
+        if send(process_dead)
+          log message
+          Process.kill(:INT, $$)
+          break
+        end
+        sleep 2
+      end
+    end
+
+    def tailwind_dead?
+      Process.waitpid(tailwind_pid, Process::WNOHANG)
+      false
+    rescue Errno::ECHILD, Errno::ESRCH
+      true
+    end
+
+    def puma_dead?
+      Process.ppid != puma_pid
+    end
+
+    def log(...)
+      log_writer.log(...)
+    end
+end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## next / unreleased
+
+* Introduce a Puma plugin to manage the Tailwind "watch" process from `rails server`. (#300) @npezza93
+
+
 ## v2.1.0 / 2023-12-19
 
 * Update to [Tailwind CSS v3.4.0](https://github.com/tailwindlabs/tailwindcss/releases/tag/v3.4.0) from v3.3.6 by @flavorjones
--- a/README.md
+++ b/README.md
@@ -60,16 +60,37 @@ The `tailwindcss:build` task is automatically attached to the `test:prepare` Rak
 
 ### Update assets automatically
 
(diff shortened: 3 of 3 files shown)


lib/puma/plugin/tailwindcss.rb
```diff
@@ -0,0 +1,65 @@
+require "puma/plugin"
+
+Puma::Plugin.create do
+  attr_reader :puma_pid, :tailwind_pid, :log_writer
+
+  def start(launcher)
+    return unless Rails.env.development?
```
brunoprietog, line 7: Wouldn't this condition be better in the Puma configuration? Only include the plugin if the environment is a development environment.

lib/puma/plugin/tailwindcss.rb
```diff
@@ -0,0 +1,65 @@
...
+
+    @log_writer = launcher.log_writer
+    @puma_pid = $$
+    @tailwind_pid = fork do
+      Thread.new { monitor_puma }
+      system(*Tailwindcss::Commands.watch_command)
+    end
+
+    launcher.events.on_stopped { stop_solid_queue }
```
brunoprietog, line 16: Ah, taken from Solid Queue. This needs to be changed

lib/puma/plugin/tailwindcss.rb
```diff
@@ -0,0 +1,65 @@
...
+
+    in_background do
+      monitor_tailwind
+    end
+  end
+
+  private
+    def stop_solid_queue
```
brunoprietog, line 24: Same here

brunoprietog (commented):
This is great! Goodbye `bin/dev`. It would be nice to change the install generator to default to include the plugin in the Puma configuration in the development environment.

lib/puma/plugin/tailwindcss.rb
npezza93: Whoops 😅

lib/puma/plugin/tailwindcss.rb
npezza93: Yea i went back and forth. I choose this one so people didnt have to worry about it in the puma config but i think having it be explicit in the puma config gives some flexibility. Changed it

npezza93:
@brunoprietog  Should have everything addressed! I also removed the bin/dev and procfile stuff. Let me know if you want to keep that around

lib/install/tailwindcss.rb
```diff
@@ -38,19 +38,8 @@
...
-
-  say "Ensure foreman is installed"
-  run "gem install foreman"
-end
-
-say "Add bin/dev to start foreman"
-copy_file "#{__dir__}/dev", "bin/dev"
-chmod "bin/dev", 0755, verbose: false
+say "Add puma plugin"
+append_to_file "config/puma.rb", %(plugin :tailwindcss if Rails.env.development?)
```
brunoprietog, line 42: Kind of nitpicky, but what do you think about adding a blank line before and after the plugin? And a brief comment saying what the plugin does. For example, Rails default has this plugin like this:

```ruby
# Allow puma to be restarted by `bin/rails restart` command.
plugin :tmp_restart
```

brunoprietog (commented):
I'm not a maintainer, I just leave my comments that I think could be useful to help this merge more quickly.

I would really love to leave bin/dev, and actuallty this same change could be very good for dartsas-rails, jsbundling-rails and cssbundling-rails if the maintainers agree.

I think it would be good to add a description to the PR. Leave bin/dev is a very convincing argument at least for me.

brunoprietog:
Nice! I have no further comments from me at least. Hopefully it will merge soon and we will have the same for the other gems that force us to use bin/dev.

flavorjones:
I think I understand why adding Puma support is valuable; but what's the rationale for deleting `lib/install/Procfile.dev` and `lib/install/dev`? That will break the developer experience for people using foreman today.

brunoprietog:
It won't break it, it would only be for new applications. `bin/dev` is horrible for debugging, and I personally don't like using foreman because it adds extra information that I don't need in the terminal. You should only need to run `bin/rails server` and that's it.

flavorjones:
There is a developer workflow that is well-documented and expected to work. Unless the files prevent users from using a puma plugin, they should not be removed. And if they do, I want to understand why.

brunoprietog:
How could you debug using `bin/dev`? Whenever I need to debug, I have to use `rails server` separately. That for me at least is quite annoying.

In my opinion, everything should just work in a single terminal. Why would you want to have two ways if one works better than the other? I don't understand why it would be better to use `bin/dev`.

I mean, why would you want to use foreman? That means adding one more dependency. It also means that in some cases, newvie developers have to use `bin/dev` and in others `rails server`. Explaining this to a newbie coming from node.js for example is not so easy, being that they can run everything in a single process there.

npezza93:
@flavorjones How would you feel about reverting all but the puma plugin and have the discussion around bin/dev in a separate PR?

flavorjones:
@npezza93 Yup, that would be great, thank you!

npezza93:
@flavorjones   Should be all set!

flavorjones:
@npezza93 Thank you! I've squashed your commits to leave a clean history, and I've reworked the README a bit to feature the Puma plugin more prominently.

This is great work, thank you! Worked great for me when I tried it out on my local dev environment.

airblade:
> I mean, why would you want to use foreman?

@brunoprietog So you can spin up other processes too, such as Redis or Sidekiq.

brunoprietog:
Redis can be started as a service just like PostgreSQL or MySQL. And Sidekiq you can also manage it with puma, so a single command is enough and you can easily debug.

hdittmar:
I tried adding the `plugin :tailwindcss...` configuration line mentioned in the Readme into my puma.rb file and since then, puma crashes on launch, for more info see here: https://github.com/rails/tailwindcss-rails/issues/319

PR #307, opened 2024-01-07
fix: password form field classes

### LINKED ISSUE #304: use scaffold generate password class is nil
use scaffold generate password class is nil

```shell
rails g scaffold user username:string password:digest
```

image

flavorjones:
Fixes #304

Previously, form would render like:

image

With this PR, form renders like:

image


dixpac:
👍 looks good to me

PR #308, opened 2024-01-07
fix: mailer generation with namespaces

### LINKED ISSUE #272: Tailwind breaks rails mailer generator for namespaced files
Rails 7.0.5.1

```
rails new mailer --css tailwind
cd mailer
rails g mailer admin/reservations_report send_csv
```

Generates error:

```
 create  app/mailers/admin/reservations_report_mailer.rb
 invoke  tailwindcss
 create    app/views/admin/reservations_report_mailer
 Could not find "layout.text.erb" in any of your source paths. Your current source paths are: 
 /Users/trueinviso/.rvm/gems/ruby-3.1.1/gems/tailwindcss-rails-2.0.29-arm64-darwin/lib/generators/tailwindcss/mailer/templates
```

Everything works as expected without the namespace, namespacing also works as expected in a new rails app without the tailwindcss-rails gem.

flavorjones:
Fixes #272

I'm a little confused why this gem implements mailer generation if it's just a copy-paste from rails, but if we're going to do it, then we need the layout files as well.

cc @dixpac from #84


dixpac:
👍

@flavorjones @dhh I wish we could just make one of the next engine versions `railties >= 7`, initially I had to jump through all kind of hoops with the generator just to make it work with both **6** and **7**, because there was a "large" change in rails 7 scaffolds.

flavorjones:
@dixpac Thanks for the :eyes: and the note about Rails 7 generators (which I wasn't aware of).

:thinking: I don't think we're currently running CI with Rails 6.x, do you think we should do that unless and until we pin to `>= 7`?

dixpac:
> 🤔 I don't think we're currently running CI with Rails 6.x, do you think we should do that unless and until we pin to >= 7?

Good question, I guess we should :)

> @dixpac Thanks for the 👀 and the note about Rails 7 generators (which I wasn't aware of).

I basically had to copy rails 7 default generator and call it tailwindcss generator, so anyone that is using rails 6 can have rails 7 like behavior with divs and partials instead of old html tables that were used in rails 6.

flavorjones:
> > 🤔 I don't think we're currently running CI with Rails 6.x, do you think we should do that unless and until we pin to >= 7?
> 
> Good question, I guess we should :)
> 
>

I'll take care of it.

PR #314, opened 2024-01-07
Allow users to override scaffold generator templates

flavorjones:
Originally reported at #164

I think this PR is preferable to #247 because it has tests and is a simpler fix.

Note that in order to test this, the test suite now uses the same temporary directory for all tests, so we are then able to set `Rails::Generators.templates_path` before the generator classes are instantiated.

cc @dixpac

### DIFF
--- a/lib/generators/tailwindcss/controller/controller_generator.rb
+++ b/lib/generators/tailwindcss/controller/controller_generator.rb
@@ -3,7 +3,7 @@ require "rails/generators/erb/controller/controller_generator"
 module Tailwindcss
   module Generators
     class ControllerGenerator < Erb::Generators::ControllerGenerator
-      source_root File.expand_path("../templates", __FILE__)
+      source_root File.expand_path("templates", __dir__)
     end
   end
 end
--- a/lib/generators/tailwindcss/controller/controller_generator.rb
+++ b/lib/generators/tailwindcss/controller/controller_generator.rb
@@ -4,6 +4,7 @@ module Tailwindcss
   module Generators
     class ControllerGenerator < Erb::Generators::ControllerGenerator
       source_root File.expand_path("templates", __dir__)
+      source_paths << "lib/templates/erb/controller"
     end
   end
 end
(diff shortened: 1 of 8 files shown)


lib/generators/tailwindcss/controller/controller_generator.rb
```diff
@@ -3,7 +3,7 @@
 module Tailwindcss
   module Generators
     class ControllerGenerator < Erb::Generators::ControllerGenerator
-      source_root File.expand_path("../templates", __FILE__)
+      source_root File.expand_path("templates", __dir__)
```
dixpac, line 6: Should we do the same here as in scaffold, i think it is something like this
`source_paths << "lib/templates/erb/controller"`

lib/generators/tailwindcss/mailer/mailer_generator.rb
```diff
@@ -3,7 +3,7 @@
 module Tailwindcss
   module Generators
     class MailerGenerator < Erb::Generators::MailerGenerator
-      source_root File.expand_path("../templates", __FILE__)
+      source_root File.expand_path("templates", __dir__)
```
dixpac, line 6: Same comment here, this will not work if someone want to override mailer templates

dixpac:
Works great with scaffolds.
However, it doesn't work with controllers or mailers. I'm not certain how frequently people would need to override those, but in those cases, it is not working.

flavorjones:
@dixpac Yeah, I was hoping nobody would care about overriding those. :laughing: :sob: 

You're right, while I'm here I might as well add mailer and controller. Thanks for the review!

lib/generators/tailwindcss/controller/controller_generator.rb
flavorjones: Fixed

lib/generators/tailwindcss/mailer/mailer_generator.rb
flavorjones: Fixed

dixpac (commented):
Awesome 😍

PR #316, opened 2024-01-08
Allow custom postcss.config.js

ahmeij:
Allows loading a custom `postcss.config.js` if it exists in the `config` folder.

This allows for instance enabling nesting like so:

```
module.exports = {
  plugins: {
    'postcss-import': {},
    'tailwindcss/nesting': {},
    tailwindcss: {},
    autoprefixer: {},
  },
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,21 +38,6 @@ This also works with relative paths. If you've installed into your app's directo
 TAILWINDCSS_INSTALL_DIR=node_modules/.bin
 ```
 
-### Using a custom postcss.config.js
-
-If you want to use a custom `postcss.config.js`, for example to enable nesting, you can place it in the `config` folder and it will be loaded automatically.
-
-```
-module.exports = {
-  plugins: {
-    'postcss-import': {},
-    'tailwindcss/nesting': {},
-    tailwindcss: {},
-    autoprefixer: {},
-  },
-}
-```
-
 ## Developing with Tailwindcss
 
 ### Configuration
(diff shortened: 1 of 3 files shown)


ahmeij:
Sorry, this looks like a duplicate of #222, although the approach is a bit different. Please let me know if this is a viable change, I can add documentation if requested.

ahmeij:
Added test coverage and updated README as was requested in the other PR.

flavorjones:
@ahmeij Thank you for opening this up! It looks good, and I appreciate the test and the documentation.

I made a small change to how the command is assembled, but if this goes green I think we should merge it.

ahmeij:
Great, thanks for the fixes and merging this change @flavorjones

muriloime:
nice work @ahmeij  ! Is there a way to run it both ways (i.e. with and without postcss) in the same app? I had opened https://github.com/rails/tailwindcss-rails/pull/222 because I need two css builds ( one for emails, without css variables, and another standard tailwindcss one).  I was wondering if I can migrate to this cleaner version

many thanks

ahmeij:
@muriloime the filenames can be made configurable, they are currently all hardcoded. 

I'd suggest a change like this, which would enable running the rake tasks multiple times for multiple files to generate.

```ruby
      def compile_command(debug: false, input_path: "app/assets/stylesheets/application.tailwind.css", output_path: "app/assets/builds/tailwind.css", config_path: "config/tailwind.config.js", tailwind_config_path: "config/postcss.config.js", **kwargs)
        command = [
          executable(**kwargs),
          "-i", Rails.root.join(input_path).to_s,
          "-o", Rails.root.join(output_path).to_s,
          "-c", Rails.root.join(config_path).to_s,
        ]

        command << "--minify" unless (debug || rails_css_compressor?)

        postcss_path = Rails.root.join(tailwind_config_path)
```

However I don't know if this is the direction @flavorjones wants to go with this. If this is ok I can make a PR like this with some tests. I would need to know how to handle the way to long list of kwargs, I think I like the following better:

```ruby
      def compile_command(debug: false, input_path: nil, output_path: nil, config_path: nil, tailwind_config_path: nil, **kwargs)
        command = [
          executable(**kwargs),
          "-i", Rails.root.join(input_path || "app/assets/stylesheets/application.tailwind.css").to_s,
          "-o", Rails.root.join(output_path || "app/assets/builds/tailwind.css").to_s,
          "-c", Rails.root.join(config_path || "config/tailwind.config.js").to_s,
        ]

        command << "--minify" unless (debug || rails_css_compressor?)

        postcss_path = Rails.root.join(tailwind_config_path || "config/postcss.config.js")
```

Issue #160, opened 2022-03-31
Rails 7.0.2.3 + Tailwind is not working, css style is not load


ardifirmansyah8:
Hi, i tried to create new rails 7 project with `--css tailwind` command. But, when i tried to run the server using `./bin/dev`, tailwind style is not load. What is i'm missing??

philipithomas:
I found the solution for my problem - it was that `config.assets.debug = true` was not set in my `development.rb` file

ardifirmansyah8:
Hi @philipithomas, i tried your solution, adding `config.assets.debug = true` in `config/environments/development.rb`, but still no luck. Do you using `tailwindcss-rails (2.0.8-x86_64-darwin)`?

ardifirmansyah8:
Ah, i found the solution, when i'm creating the controller, i'm extend `ActionController` not `ApplicationController`, ex: `class HomeController < ApplicationController`.

JesusGautamah:
i think the problem was the command, generate it with --css tailwindcss instead

Swalden:
I ran into a similar issue with some classes not loading. Adding config.assets.debug = true fixed the issue.

is config.assets.debug = true required for tailwind CSS to function locally? If so should it be included in the readme?

Khondaker97:
I am facing the same issue. Adding config.assets.debug = true hasn't done anything.

jmtame:
Thanks @Khondaker97 you saved me some time!

JesusGautamah:
> I was facing the same issue. Adding config.assets.debug = true hasn't done anything. Then, rake tailwindcss:build command in bash solved my issue.

What's about 
```
tailwindcss:watch
```

and 

```
rake assets:precompile
```

therealrodk:
Adding `config.assets.debug = true` to `development.rb` fixed my issue, as well. Can anyone explain why this is necessary? I'm a little confused, especially since it was working properly and then suddenly stopped applying new styles I added.

kokorolx:
`config.assets.debug = true` fixed my issue. Thank you

rails 7
ruby 3.2.2
MacOS: 13.2.1
Chipset: M1

dhh:
I can't replicate these issues on Rails 7.0.5 with tailwind-rails 2.0.29. TW styles load as they should. There's no `assets.debug = true` necessary. Running using ./bin/dev.

Bishoy-Samwel:
I have the same issue 
with ruby '3.2.1'
[...]

PR #321, opened 2024-01-21
Update docs to show required config

Michael9311:
See https://github.com/rails/tailwindcss-rails/issues/160

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,7 +88,6 @@ If you are running `rails tailwindcss:watch` in a docker container without a tty
 
 Additionally, you must ensure that `config.assets.debug = true` is set in `development.rb`.
 
-
 #### Foreman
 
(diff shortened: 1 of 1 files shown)


flavorjones:
As mentioned by @dhh [here](https://github.com/rails/tailwindcss-rails/issues/160#issuecomment-1596102836), the maintainers are unable to reproduce what is being described at #160, and this documentation is incorrect.

If you can help us reproduce what's happening, please open an issue so we can fix it! Thank you!

Z2Flow:
I had the same problem and the solution by @Michael9311 solved the issue.
Rails 7.1.3.2
Ruby 3.3.0
tailwindcss-rails 2.3.0

I created a simple Rails project:

`rails new Play -css tailwind `
`rails g scaffold Post title:string body:text`

PR #331, opened 2024-02-08
Install as Puma plugin if available

javierav:
This PR implements a first approach to resolve #327. The README remains to be updated if this solution is finally accepted.

### DIFF
--- a/lib/install/tailwindcss.rb
+++ b/lib/install/tailwindcss.rb
@@ -38,19 +38,24 @@ unless Rails.root.join("app/assets/stylesheets/application.tailwind.css").exist?
   copy_file "#{__dir__}/application.tailwind.css", "app/assets/stylesheets/application.tailwind.css"
 end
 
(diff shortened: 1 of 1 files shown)


flavorjones:
Thank you for writing this up! It seems reasonable, but I do have a concern that we're doing a lot to try to guess (and then not break) the user's development workflow.

I have a slight preference for #347 which is more general, if we can figure out how to land it. Let's leave this open for now, but I'll circle back once we made a decision on the other PR.

flavorjones:
Circling back on this. The other PR was a good approach but DHH opined that it should be done in Rails (not in an individual asset gem like this).

This PR's change is much simpler, but a) running `bin/dev` will kick off *two* build watch processes (one by the Procfile and one by the Puma plugin), and b) it should arguably still be moved into Rails so that it's not living in individual asset gems.

WDYT? If you want to try to address (a) then we can discuss (b) with David.

PR #349, opened 2024-04-24
Avoid stdin read by tailwindcss watch command

tompng:
## Description
`tailwindcss -w` reads stdin, probably to detect stdin close.
When using binding.irb or debugger with tailwindcss-rails, some keystrokes are taken by tailwindcss  and IRB/Debug can't read from stdin correctly. Keystroke disappears, and sometimes hangs up.
To not let tailwindcss watch command read stdin, we should use `IO.popen(command, 'r+')` instead of `system(*command)`.
Workaround for https://github.com/rails/tailwindcss-rails/discussions/346

## Reproduction code
```ruby
watch_command = ['./node_modules/.bin/tailwindcss', '-w']
Thread.new do # same for `fork do`
  system(*watch_command)
  # IO.popen(watch_command, 'r+') do |io|
  #   IO.copy_stream(io, $stdout)
  # end
  puts 'command finished'
end
sleep 1
binding.irb
```

## Side effect
Before this pull request, Ctrl-D will stop `rails s`. Closing stdin will stop tailwind, then puma will stop with a message `Detected tailwind has gone away, stopping Puma...`
After this change, Ctrl-D won't stop tailwind.

### DIFF
--- a/lib/puma/plugin/tailwindcss.rb
+++ b/lib/puma/plugin/tailwindcss.rb
@@ -8,7 +8,12 @@ Puma::Plugin.create do
     @puma_pid = $$
     @tailwind_pid = fork do
       Thread.new { monitor_puma }
-      system(*Tailwindcss::Commands.watch_command)
+      # Using IO.popen(command, 'r+') will avoid watch_command read from $stdin.
+      # If we use system(*command) instead, IRB and Debug can't read from $stdin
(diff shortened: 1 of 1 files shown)


flavorjones:
@tompng Thank you for opening this! Will try it out.

flavorjones:
Shipped in https://github.com/rails/tailwindcss-rails/releases/tag/v2.4.1

PR #357, opened 2024-04-29
generator: sync erb template

### LINKED ISSUE #341: Tailwind scaffold template is out of sync with rails/rails ERB scaffold template
https://github.com/rails/tailwindcss-rails/blob/main/lib/generators/tailwindcss/scaffold/templates/index.html.erb.tt is out of sync with https://github.com/rails/rails/blob/main/railties/lib/rails/generators/erb/scaffold/templates/index.html.erb.tt, namely it's missing the `<% content_for :title, "New event" %>`

Should template files be kept in sync (except the tailwind classes)?

Screenshot 2024-03-24 at 21 55 11

kinsomicrote:
This updates the ERB template to be in sync with the template in rails/rails

Fixes #341.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## unreleased
+
+* Bring the scaffold templates up to date with rails/rails as much as possible without breaking 6.1 compatibility. (#357) @kinsomicrote
+
+
 ## v2.5.0 / 2024-04-27
 
 * Remove the `@tailwindcss/aspect-ratio` plugin from the `tailwind.config.js` that gets installed by the generator. This plugin was originally a polyfill until Safari 15 was released (in Fall 2021), and so is beyond its useful lifetime for anyone not targetting ancient browsers. (#344) @flavorjones @searls
--- a/lib/generators/tailwindcss/scaffold/templates/edit.html.erb.tt
+++ b/lib/generators/tailwindcss/scaffold/templates/edit.html.erb.tt
@@ -3,6 +3,6 @@
 
   <%%= render "form", <%= singular_table_name %>: @<%= singular_table_name %> %>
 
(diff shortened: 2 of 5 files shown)


kinsomicrote:
Hello @flavorjones 👋 

Any pointers on why the test is failing? I ran the test suite locally, and they all passed.

javierjulio:
@kinsomicrote based on the errors occurring for the Rails 6.1 build, it looks like `model_resource_name` and `index_helper` have a different method signature at that version. So using `model_resource_name` as an example, [the v7.1 docs](https://api.rubyonrails.org/classes/Rails/Generators/NamedBase.html#method-i-model_resource_name) differ from [the v6.1 docs](https://api.rubyonrails.org/v6.1.0/classes/Rails/Generators/NamedBase.html#method-i-model_resource_name).

flavorjones (changes requested):
@kinsomicrote Thank you for submitting this! I'd prefer to use a calling convention that's compatible with Rails 6.1 (rather than introduce version checks).

kinsomicrote:
@flavorjones tests are passing now. Let me know if it's good to go.

flavorjones:
Quick note for posterity: the `model_resource_name` change is from rails/rails#43611 which first appeared in Rails 7.0 and fixed namespaced models.

Namespaced models (e.g. `Admin::Monster`) still don't work with this PR but that's probably OK as long as we have to support Rails 6.1

flavorjones:
I've rebased and squashed

flavorjones:
@kinsomicrote Thanks so much for doing this work!!!

I'm going to open a second PR with your changes using the Rails 7.0 API so we can have a separate conversation about dropping Rails 6.1 feature support.

PR #359, opened 2024-05-01
generator: sync erb template to use Rails 7.0 APIs

flavorjones:
This PR is derived from the work @kinsomicrote did in #357.

Updates the erb templates to be in sync with the templates in rails/rails 7.0+ and so correctly supports namespaced models (see related rails/rails#43611).


flavorjones:
@dhh Are you comfortable with dropping Rails 6.1 feature support in a minor release (likely v2.7.0) if/when this gets merged?

PR #388, opened 2024-08-06
Disable loading .env file when running foreman for development

Flixt:
Same as in rails/cssbundling-rails#155 and rails/jsbundling-rails@0ff85a2255b732556ee30d28634ecdb15a50d0ae

We should just not load the .env when running foreman locally for bundling etc.

### DIFF
--- a/lib/install/dev
+++ b/lib/install/dev
@@ -13,4 +13,4 @@ export PORT="${PORT:-3000}"
 export RUBY_DEBUG_OPEN="true"
 export RUBY_DEBUG_LAZY="true"
 
-exec foreman start -f Procfile.dev "$@"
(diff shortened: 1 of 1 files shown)


dhh:
We're actually fixing this in Kamal 2, which will be out before Rails 8, so that it uses .kamal/secrets instead of .env. So I'm not sure this is actually necessary, and would prevent the use of .env for legitimately passing on envs to foreman.

Flixt:
Okay. But I think that tailwindcss-rails, cssbundling-rails, jsbunding-rails and dartsass-rails should generate a `bin/dev` file with the same behavior. Currently only cssbundling-rails and jsbundling-rails generate `bin/dev`  excluding the env.

PR #406, opened 2024-09-25
Extract tailwindcss packaging into an external gem, `tailwindcss-ruby`

flavorjones:
In advance of the upcoming TailwindCSS v4 release, we are decoupling the `tailwindcss` executable from the Rails integration. This will allow users to upgrade TailwindCSS at a time of their choosing, and allow early adopters to start using the beta releases.

See https://github.com/flavorjones/tailwindcss-ruby


Gemfile.lock
```diff
@@ -128,6 +129,7 @@ GEM
       io-console (~> 0.5)
     securerandom (0.3.1)
     stringio (3.1.1)
+    tailwindcss-ruby (0.1.0-x86_64-linux)
```
skatkov, line 132: **issue:** I assume that this change would not work as is.

0.1 version of `tailwindcss-ruby` gem doesn't contain this fix:
https://github.com/flavorjones/tailwindcss-ruby/commit/bd43725db3a328962d04a4c43d9ce4552c45e8c1

flavorjones: Thanks for the note. I'm still bootstrapping the gem! Obviously everything will get fixed before a real release. ❤️

skatkov: So looking forward to that!

I want to use it for Jekyll integration as well ;)

flavorjones:
@dhh I think this is ready to go, if you're OK with it.

Gemfile.lock
flavorjones: Yes! I've updated my personal site (jekyll-based) to use this gem and it's totally working for me.

https://github.com/flavorjones/flavorjones.github.io/commit/03b9acea1daeea915a9507c3f81761dc4f1b7dc5

skatkov: @flavorjones amazing work! Thank you

`jekyll-tailwind-cli` gem is just 4 lines of code now thanks to you. I might as well just scrap it. :)

dhh (approved):

flavorjones:
This was shipped in https://github.com/rails/tailwindcss-rails/releases/tag/v3.0.0.rc1, please try it out and [let me know if you have any problems](https://github.com/rails/tailwindcss-rails/discussions/411).

PR #407, opened 2024-09-29
Rename Authentication template files

seanpdoyle:
Related to https://github.com/rails/rails/pull/53103
Follow-up to [rails/rails#53098][]

The corresponding commit in `rails/rails` nested the template files (including the view templates) under an `app/` directory.

This commit makes a corresponding change, nesting the `views/` templates under `app/views/`.

[rails/rails#53098]: https://github.com/rails/rails/pull/53098#issuecomment-2381650236


flavorjones:
@guilleiguaran What's your plan for making sure this is backwards compatible with Rails 8 beta1? Currently this change breaks with Rails 8 beta1, which means I can't cut a release of this gem until there's a beta2 with the above change.

guilleiguaran:
@flavorjones we can revert this and cut the release of the gem and then re-apply to ensure this is in place for the beta2 compatible release.

flavorjones:
I'll figure it out.

flavorjones:
See https://github.com/rails/tailwindcss-rails/pull/408 and https://github.com/rails/tailwindcss-rails/pull/409

PR #409, opened 2024-10-01
Revert auth templates for Rails 8 beta1

flavorjones:
Do not merge until Rails 8 beta2 is released.

Reverts part of #408 which fixed #407's backwards-incompatible change.


flavorjones:
The failure on `upstream / user-journey (rails v8.0.0.beta1)` is expected; once beta2 is out I'll bump the matrix and this should go green.

PR #416, opened 2024-10-08
Fix naming regarding Rails form naming

enderahmetyurt:
Fix #415

PR #418, opened 2024-10-09
Handle both rails 7.2 and rails 8.0 field types for checkbox/textarea

Earlopain:
Followup to https://github.com/rails/tailwindcss-rails/pull/416

They got renamed (with aliases being present) in:
* https://github.com/rails/rails/pull/52432
* https://github.com/rails/rails/pull/52467

This ensures the correct classes are generated for both old and new versions. I added a test locally but it can only assert against the exact css classes since there is nothing else to properly distinquish them. I can add this here if you wish, just didn't seem like the best test.

@enderahmetyurt, I suspect you generated your sample app with rails 8.0.0.beta1 which is why it looked weird for you.


enderahmetyurt:
Good job @Earlopain yes I created rails 8.0.0.beta1

skipkayhil (approved):

flavorjones:
@Earlopain Thank you for catching this!

PR #420, opened 2024-10-17
Add experimental support for Tailwind CSS v4

### LINKED ISSUE #419: tailwind.css file is not being updated in Rails 8 with tailwindcss-ruby 4.0.0.alpha27
Whenever I update the `app/assets/stylesheets/application.tailwind.css` file, the corresponding output file at `app/assets/builds/tailwind.css` does not update.

The process seems to be doing something:

image

But the changes are not being picked up. Notice the execution time on the watcher vs the updated timestamps on the file:

Screenshot 2024-10-17 at 11 21 24 PM

The only way to get this to work is to delete the file, stop the watcher (i.e. `bin/dev` in a rails app) and run the watcher again.

I created an app to try out the new setup and experiment on Tailwind v4. The only notable thing about it is that I removed the `tailwind.config.js` file since it's not supposed to be needed in Tailwind v4. I tried adding it back and that did not resolve the issue.

https://github.com/pinzonjulian/tw4-experiments

**Gem versions**
tailwindcss-rails (3.0.0)
tailwindcss-ruby (4.0.0.alpha.27)
rails (8.0.0.beta1)

Running on macOS Sonoma 14.6.1

flavorjones:
Tailwind CSS v4 is a big overhaul in how tailwind detects CSS classes. Instead of specifying config and input files, the CLI scans the entire project directory.

More information on this change here: https://tailwindcss.com/blog/tailwindcss-v4-alpha

This PR updates the build and watch commands to skip the `-i` and `-c` parameters if the version of Tailwindcss::Ruby is v4.

This PR also adds matrixed (across Tailwind v3 and v4) unit test coverage for the base commands, and integration test coverage for the generated CSS file.

Known issue: this logic will break for users who are using the `TAILWINDCSS_INSTALL_DIR` feature of `tailwindcss-ruby` to run a different version of tailwindcss than declared by the gem. This PR hides that version check behind the `Commands.tailwindcss_version` method, so we can deal with that once someone complains about it.

Closes #419

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,7 +37,10 @@ jobs:
       fail-fast: false
       matrix:
         plat: ["ubuntu", "windows", "macos"]
+        tailwind: ["--version=~>3.4.14", "--version=~>4.0.0.alpha.27"]
+    env:
     runs-on: ${{matrix.plat}}-latest
+      TAILWINDCSSOPTS: ${{ matrix.tailwind }}
     steps:
       - uses: actions/checkout@v4
       - uses: ruby/setup-ruby@v1
--- a/.github/workflows/upstream.yml
+++ b/.github/workflows/upstream.yml
@@ -42,8 +42,10 @@ jobs:
       matrix:
         plat: ["ubuntu"]
         ref: ["7-2-stable", "v8.0.0.beta1", "main"]
+        tailwind: ["--version=~>3.4.14", "--version=~>4.0.0.alpha.27"]
     env:
       RAILSOPTS: --git=https://github.com/rails/rails --ref=${{ matrix.ref }}
+      TAILWINDCSSOPTS: ${{ matrix.tailwind }}
     steps:
       - uses: actions/checkout@v4
       - uses: ruby/setup-ruby@v1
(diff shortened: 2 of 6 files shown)


pinzonjulian:
Hey Mike

Unfortunately this didn't work and I believe it's even introduced some regressions. Before I could use Tailwind 4's css-style config but not that isn't working. Something like this has no effect in the output file. 

```css
/* app/assets/stylesheets/application.tailwind.css */
@import "tailwindcss";

@theme {
  --color-*: initial;
  --shadow-*: initial;
  --inset-shadow-*: initial;
  --drop-shadow-*: initial;

  --color-special: #fafafa;
}
```

Before this change it did create an output file with the intended variables (as mentioned before I had to delete and rebuild but it worked). Now it just outputs the whole of Tailwind without any modifications. I even tried commenting out all of the file or adding just the preflight has no effect; I always get the full, default tailwind installation:

```css
/* app/assets/stylesheets/application.tailwind.css */

@import "tailwindcss/preflight" layer(base);
```

flavorjones:
Let's take the conversation back to #419 please since I've obviously misunderstood your use case.

PR #424, opened 2024-10-18
Add failing spec for Tailwind v4

pinzonjulian:
This PR adds a failing assertion to the integration test. The new test:
- Adds a `@theme` block 
- inserts a custom property
- runs the build command
- asserts that the custom property is in the output file.

I'm completely new to the codebase so there's probably a lot that's missing here but I wanted to start the conversation with a failing spec to get things going.


flavorjones:
Thank you for this! This is essentially the same test I wrote in #423 so that PR includes this coverage and fixes it. I'm glad we validated that.

PR #451, opened 2025-01-05
Update README.md

mackross:


flavorjones:
@ahmeij In your opinion, is this valuable to add to the readme? It seems obvious to me but may not be obvious to some users.

ahmeij:
For me, if I did not know it would not help me enough to solve the problem, as in, what package manager and would i need to  do. 

Maybe something like this:

Also a `package.json` with dependencies for your plugins and a package manager like `yarn` or `npm` is required:

```json
// package.json
{
  "name": "my app",
  "private": true,
  "dependencies": {
    "postcss-advanced-variables": "^4.0.0",
    "postcss-import": "^16.0.1",
    "postcss-mixins": "^9.0.4",
    "tailwindcss": "^3.4.1"
  }
}
```

Use yarn or npm to install the dependencies.

flavorjones:
@ahmeij Thanks for your advice. I've committed 32d4da09 made you co-author.

@mackross I'm going to close this in favor of the above suggestion in that commit, but thank you for pushing us to clarify the docs.

PR #454, opened 2025-01-10
Improve boolean fields

patriciomacadden:
In this opportunity I'm changing the boolean fields slightly:

* align the checkbox and the label in the form
* display `Yes` and `No` instead of `true` or `false`

| before | after |
| --- | --- |
| checkbox-before | checkbox-after |
| show-before | show-after |


flavorjones:
@patriciomacadden It looks like there's one other notable change in this PR, which is to assign text fields `order-first`, which I'm not necessarily opposed to! But I wanted to check that it is here intentionally.

I also noticed some reordering of attributes, which I'm very much in favor of. Do you want to submit a PR with just that being done, as a "cleanup" PR?

patriciomacadden:
Apologies @flavorjones , I messed up with the class reordering. turned it off now. fyi, the vscode extension i'm using for this purpose (to keep consistency across different projects is [headwind](https://marketplace.visualstudio.com/items?itemName=heybourn.headwind)). More than happy to contribute a pr with class reordering but i don't think it makes a lot of sense as we can't enforce people to use it... 

Regarding the `order-first` class, it's only applied to attributes that are either `:checkbox` or `:check_box`.

Essentially, what i changed is:
* add the flex class if it's a `:check_box` or `:checkbox` field so the checkbox and the label are aligned and not displayed one below the other
* add `order-first` to the input (when it's a `:checbox` or `:check_box`) so the checkbox is rendered on the left.

did i miss something?

flavorjones:

patriciomacadden:
just rebased, squashed and fixed some class names changes from upgrading to v4.

PR #457, opened 2025-01-17
make the index a table again

patriciomacadden:
As per [this PR](https://github.com/rails/rails/pull/41210), the scaffold generator stopped generating a table in the index page, and encourages css frameworks gems (such as this) to do the styling. 

I don't quite like what the generator generates currently (design wise, although I'm no designer), so I thought of bringing back the table, with a basic style so it's ready to use (or as ready to use as possible).

I looked for box-style CRUD designs so we could keep using the partial but I couldn't find anything.

As always, before and after pictures:

| before | after |
| --- | --- |
| image | image |

### DIFF
--- a/lib/generators/tailwindcss/scaffold/templates/index.html.erb.tt
+++ b/lib/generators/tailwindcss/scaffold/templates/index.html.erb.tt
@@ -2,11 +2,11 @@
 
 <div class="w-full">
   <%% if notice.present? %>
-    <p class="inline-block px-3 py-2 mb-5 font-medium text-green-500 rounded-md bg-green-50" id="notice"><%%= notice %></p>
+    <p class="py-2 px-3 bg-green-50 mb-5 text-green-500 font-medium rounded-md inline-block" id="notice"><%%= notice %></p>
   <%% end %>
 
-  <div class="flex items-center justify-between">
-    <h1 class="text-4xl font-bold"><%= human_name.pluralize %></h1>
+  <div class="flex justify-between items-center">
+    <h1 class="font-bold text-4xl"><%= human_name.pluralize %></h1>
     <%%= link_to "New <%= human_name.downcase %>", new_<%= singular_route_name %>_path, class: "rounded-md px-3.5 py-2.5 bg-blue-600 hover:bg-blue-500 text-white block font-medium" %>
   </div>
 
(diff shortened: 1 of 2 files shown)


lib/generators/tailwindcss/scaffold/templates/index.html.erb.tt
```diff
@@ -2,24 +2,40 @@
...
+        <tbody class="divide-y divide-gray-200">
+          <%% @users.each do |user| %>
+            <tr class="border-b">
+<% attributes.reject(&:password_digest?).each do |attribute| -%>
+              <td class="py-4 whitespace-nowrap"><%%= <%= singular_name %>.<%= attribute.column_name %> %></td>
+<% end -%>
+              <td class="py-4 space-x-4 whitespace-nowrap">
+                <%%= link_to "Show", <%= model_resource_name(singular_table_name) %>, class: "text-blue-600 hover:text-blue-900" %>
+              </td>
+            </tr>
```
patriciomacadden, line 33: we can still have this in a partial but i don't see the benefit - i don't think it'll be used somewhere else.

lib/generators/tailwindcss/scaffold/templates/index.html.erb.tt
```diff
@@ -2,24 +2,40 @@
...
+          </tr>
+        </thead>
+        <tbody class="divide-y divide-gray-200">
+          <%% @users.each do |user| %>
+            <tr class="border-b">
+<% attributes.reject(&:password_digest?).each do |attribute| -%>
+              <td class="py-4 whitespace-nowrap"><%%= <%= singular_name %>.<%= attribute.column_name %> %></td>
+<% end -%>
+              <td class="py-4 space-x-4 whitespace-nowrap">
+                <%%= link_to "Show", <%= model_resource_name(singular_table_name) %>, class: "text-blue-600 hover:text-blue-900" %>
```
patriciomacadden, line 31: I kept just the link to the show action, but we can also add the Edit and Destroy links. thoughts?

flavorjones: I would personally like to have the Edit and Destroy links in the table, but I suppose we could do that in a separate PR if you're unsure.

lib/generators/tailwindcss/scaffold/templates/index.html.erb.tt
```diff
@@ -2,24 +2,40 @@
...
+        <thead>
+          <tr>
+<% attributes.reject(&:password_digest?).each do |attribute| -%>
+            <th class="py-3.5 text-left font-semibold"><%= attribute.human_name %></th>
+<% end -%>
+            <th class="py-3.5 text-left font-semibold">Actions</th>
+          </tr>
+        </thead>
+        <tbody class="divide-y divide-gray-200">
+          <%% @users.each do |user| %>
```
flavorjones, line 25: This needs to be `@<%= plural_table_name %>.each`

lib/generators/tailwindcss/scaffold/templates/index.html.erb.tt
```diff
@@ -2,24 +2,40 @@
...
+            <tr class="border-b">
+<% attributes.reject(&:password_digest?).each do |attribute| -%>
+              <td class="py-4 whitespace-nowrap"><%%= <%= singular_name %>.<%= attribute.column_name %> %></td>
```
flavorjones, line 28: The partial has logic to handle rendering links to attachments, which we should probably mirror.

lib/generators/tailwindcss/scaffold/templates/index.html.erb.tt
flavorjones: I think it's fine to have it inline in this PR, if a need emerges for a partial it's easy to extract later.

flavorjones (changes requested):

lib/generators/tailwindcss/scaffold/templates/index.html.erb.tt
patriciomacadden: 🤦

lib/generators/tailwindcss/scaffold/templates/index.html.erb.tt
patriciomacadden: good catch - added!

lib/generators/tailwindcss/scaffold/templates/index.html.erb.tt
patriciomacadden: I think it's a nice addition - added!

lib/generators/tailwindcss/scaffold/templates/index.html.erb.tt
patriciomacadden: perfect!

patriciomacadden:
thanks for your review @flavorjones, all issues addressed! :)

patriciomacadden:
I'm having second thoughts on converting to a table. I do think it's the proper way to implement a CRUD (and it's always been in Rails) but now it's not (see [this comment](https://github.com/rails/tailwindcss-rails/pull/71#issuecomment-912960884) in the PR that originated the current design). 

But also I'd like the scaffold generator to generate something ready to use without too many modifications.

I'm closing this PR, and I'll see how can we improve the list view.

Feel free to reopen if you think it's good, though.

PR #460, opened 2025-01-20
improve partial: correct use of tags + attachment links styling

patriciomacadden:
The partial used `<p>` elements for each attribute, which auto-close if it contains a `<div>` element (see [the technical summary of the p tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/p#technical_summary)). To avoid this, I'm changing the `p` elements to `div` elements.

Also, adding some basic styling for the attachment links.

See before and after screenshots:

| before | after | comment |
| --- | --- | --- |
| image | image | the space between "Documents" and the links changed due to the correct use of tags, and the links now are styled differently (similar to "forgot your password?" link |
| image | image | the p element auto-closed due to the inner div, by changing it to a div we produce a better output |


flavorjones:

PR #464, opened 2025-01-23
Introduce proper v4 support for new apps

### LINKED ISSUE #463: Installation should assume Tailwind CSS v4
When running "rails tailwindcss:install", the resulting application does not apply styles with `tailwindcss-ruby` v4.0.0.

I think what needs to happen is:

- bump dependency to `tailwindcss-ruby ~> 4.0` in the gemspec
- drop the config file and move it into `application.tailwind.css`
- add anything that the `@tailwindcss/upgrade@next` injects into `application.tailwind.css` for any of our generated views
- drop the Inter font (see #353)
- bump the major version of this gem to v4 (a happy coincidence)

flavorjones:
Main changes to the installer:

- bump the dep on tailwindcss-ruby to `~> 4.0`
- drop the `config.tailwind.js` file, and the `-c` option from the tailwindcss command
- add `import "tailwindcss";` to `application.tailwind.css`
- update class names in the templates
- remove the Inter font (see below)

Note that this PR also removes the Inter font, which may leave an orphaned stylesheet reference in existing apps that upgrade:

```
# previously installed into layouts/application.html.erb
<%= stylesheet_link_tag "tailwind", "inter-font", "data-turbo-track": "reload" %>
```

This will need to be replaced/fixed by an upgrade generator/task, but I don't think it needs to be part of this PR right now.

cc @EricGusmao for feedback

Closes #463


EricGusmao:
@flavorjones wow, so fast! There is still something missing though, renaming the generator class names to suport the newer ones that were renamed, as my commit https://github.com/rails/tailwindcss-rails/pull/465/commits/fd342bc48f66ef1e124c75ef8d82d9bdc088319a

for reference: https://tailwindcss.com/docs/upgrade-guide#renamed-utilities

flavorjones:
Edited this PR to remove the injected compatibility styles. I'll make that part of the upgrade generator in a separate PR.

flavorjones:
@EricGusmao Thanks for the commit updating class names, I've cherry-picked that into this PR.

EricGusmao:
> Edited this PR to remove the injected compatibility styles. I'll make that part of the upgrade generator in a separate PR.

so the idea is to create a upgrade generator to patch current installations?

flavorjones:
> so the idea is to create a upgrade generator to patch current installations

Yes, I've got one ready to create a PR with that's pretty simple -- does minimal patching and then runs the upstream upgrader.

@EricGusmao Do you agree this PR is OK to merge? Can you think of anything missing for the "tailwindcss:install" use case in a new app?

flavorjones:
I think this is a good improvement, so I'm going to merge it. We can continue to fix things as we find them.

EricGusmao:
> > so the idea is to create a upgrade generator to patch current installations
> 
> Yes, I've got one ready to create a PR with that's pretty simple -- does minimal patching and then runs the upstream upgrader.
> 
> @EricGusmao Do you agree this PR is OK to merge? Can you think of anything missing for the "tailwindcss:install" use case in a new app?

yes, lgtm

Issue #462, opened 2025-01-23
Improve support for Tailwind CSS v4


flavorjones:
**Please do not ask for advice about v4 here.**

**Please go to [TailwindCSS v4 - upgrade experience report · Discussion #450](https://github.com/rails/tailwindcss-rails/discussions/450) for help, questions, and discussion.**

---

This issue is intended to drive concrete planning for better v4 support.

The state of the gem at the moment:

- tailwind css v4 was released 2025-01-23, and tailwindcss-ruby also released 4.0.0 the same day
- there is an upgrade process that existing applications should go through
  - [x] we should tell people to stick on 3.x until they're ready to upgrade
  - [x] we should document some steps to help people do this
    - https://github.com/rails/tailwindcss-rails/pull/467
  - [x] if possible, we should automate the upgrade process with a generator/task
    - https://github.com/rails/tailwindcss-rails/pull/466 
- new rails applications don't seem to work properly
  - [x] #463

What else do we need to do?

Notably, I say "we" because I'm asking for help with this. I haven't upgraded a full app to v4 myself and so people who have should feel obligated to help get the gem into shape. Thanks in advance.

flavorjones:
Closing. Shipped a release candidate: [Release v4.0.0.rc1 / 2025-01-23 · rails/tailwindcss-rails](https://github.com/rails/tailwindcss-rails/releases/tag/v4.0.0.rc1)

PR #466, opened 2025-01-23
Introduce a 'tailwindcss:upgrade' task

flavorjones:
- rename install/tailwindcss.rb to install/install_tailwindcss.rb
- new script install/upgrade_tailwindcss.rb which:
  - removes the "inter-font" CSS tag from the application layout
  - runs "npx @tailwindcss/upgrade@next --force" if npx is available
- new integration test for upgrade

Child of #462


EricGusmao:
As I mentioned in [#450 ](https://github.com/rails/tailwindcss-rails/discussions/450#discussioncomment-11930704), the update tool assumes `postcss.config.js`  is at the current directory where it is executed.

And I noticed  this line wasn't changed https://github.com/rails/tailwindcss-rails/blob/629da535f41cd29857d60c77d470a779c1f2f70d/lib/tailwindcss/commands.rb#L17 

Maybe we should change the update task to also move the postcss config file to the root path

change the previous line to

```ruby
postcss_path = rails_root.join("postcss.config.js") 
```

And update the documentation related to postcss to reflect accordingly. 

What do you think about this approach? Are there any potential drawbacks or edge cases I might have missed?

flavorjones:
@EricGusmao good catch. I'll add that to this PR.

flavorjones:
@EricGusmao added a commit to handle the postcss file.

EricGusmao:
@flavorjones There’s one aspect that still concerns me a bit. If the user isn’t using PostCSS, some migrations (e.g., updating class names in the view files) will fail.

Additionally, in setups without JavaScript tooling, the update process may fail to fully migrate ``tailwind.config.js`` because the tool assumes that the imported packages (e.g., tailwind plugins) are installed via a package manager, allowing them to be called.

Do you think it’s worth addressing these scenarios?

flavorjones:
@EricGusmao If I knew how to address those scenarios, I might be willing to try. But I think there is a limit to how much we can reasonably automate, and the users will have to do some manual work.

In other words, I think this is "close enough to ship", and we can continue to make the upgrade task better if we figure out ways to address edge cases like the ones you're mentioning.

Is that a reasonable response?

EricGusmao:
@flavorjones Yes, I completely agree.

PR #412, opened 2024-10-02
Prevent duplicate tailwind css reference

rip747:
On a newly generated 8.0.0.beta1 app it seems that the tailwind.css is referenced twice with the head of the generated layout

```
<link rel="stylesheet" href="/assets/tailwind-aaa98186.css" data-turbo-track="reload" />
<link rel="stylesheet" href="/assets/inter-font-1d90a608.css" data-turbo-track="reload" />
<link rel="stylesheet" href="/assets/application-8b441ae0.css" data-turbo-track="reload" />
<link rel="stylesheet" href="/assets/application.tailwind-bb2039bf.css" data-turbo-track="reload" />
<link rel="stylesheet" href="/assets/tailwind-aaa98186.css" data-turbo-track="reload" />
```

it seems that the injected stylesheet_link_tag is including the tailwind stylesheet 
```
<%= stylesheet_link_tag "tailwind", "inter-font", "data-turbo-track": "reload" %>
```

it is automatically included from the 'app' stylesheet call from the app/build directory

`<%= stylesheet_link_tag :app, "data-turbo-track": "reload" %>`

removing the tailwind reference 
```
<%= stylesheet_link_tag "inter-font", "data-turbo-track": "reload" %>
```

prevents the duplicate in the generated layout

```
<link rel="stylesheet" href="/assets/inter-font-1d90a608.css" data-turbo-track="reload" />
<link rel="stylesheet" href="/assets/application-8b441ae0.css" data-turbo-track="reload" />
<link rel="stylesheet" href="/assets/application.tailwind-bb2039bf.css" data-turbo-track="reload" />
<link rel="stylesheet" href="/assets/tailwind-aaa98186.css" data-turbo-track="reload" />
```

PR #473, opened 2025-01-24
fix: unnecessary stylesheet_link_tag for tailwind

flavorjones:
In Rails 8, propshaft detects `:app` and loads all the css files under app/assets/stylesheets, so we may not need this explicit tag for tailwind.

On installation, omit the tailwind stylesheet link tag if `:app` is already being passed to stylesheet_link_tag in the layout.

On upgrade, remove the tailwind stylesheet link tag if `:app` is already being passed to stylesheet_link_tag in the layout.

See related discussion at https://github.com/rails/tailwindcss-rails/pull/412


brunoprietog:
This doesn't seem to be the behavior we need, since, if you use Tailwind, the `:app` argument will include all *.css files inside app/assets. Here we are only interested in `tailwind`, which is the final build of all the CSS files you might have if you use `@import`.

With this, we'd be including default CSS files that have tailwind syntax that wouldn't be appropriate.

brunoprietog:
Ignore me, just saw #474 👍

PR #474, opened 2025-01-24
Move application.tailwind.css to a dir ignored by propshaft

### LINKED ISSUE #470: v4.0.0-rc1 - Bundled application.tailwind.css gets 404 trying to load /assets/tailwindcss
Hi, thank you for making this library. It's been awesome to get setup.

I have everything working, and it looks great, but I noticed in my Chrome Dev Tools Console an error:

Image

The error doesn't seem to be breaking anything, but I'm not sure what's triggering it. It seems that the bundled version of application.tailwind.css, which just has the contents `@import "tailwindcss";` is triggering a network call to GET http://localhost:3000/assets/tailwindcss. I'm not super familiar with the @import statement in CSS, but I find that odd.

The Sources tab shows:

Image

If I comment out that import in application.tailwind.css in an attempt to stop the error, Tailwind stops loading all together and my page formatting goes ugly.

This is a pretty "stock" v4.0.0-rc1 install. Screenshots of my diff below.

<hr>

Image

<hr>

Image

<hr>

tailwind.config.js deleted, because I basically got to 4.0.0-rc1 but removing everything from 3.x and running the `tailwindcss:install` task with 4.x which [doesn't install tailwind.config.js](https://github.com/rails/tailwindcss-rails/blob/65dd6c5ef76e474bf73249a086437bafa58e49b4/CHANGELOG.md?plain=1#L14C8-L14C27).

Image

flavorjones:
This file should not be included in the stylesheet link tags served to user agents. But by default, Rails 8 apps add a tag for all CSS files under `app/assets` (courtesy of a Propshaft feature handling a source named `:all`), and so this file is being included unintentionally.

Going forward, in version 4.x of this gem (which pins to Tailwind v4), the installer will place the file into `app/assets/tailwind`. The upgrader will move it from `app/assets/stylesheets` to `app/assets/tailwind`.

And the gem's railtie adds `app/assets/tailwind` to `config.assets.excluded_paths` to make sure it doesn't get served up by the `stylesheet_link_tags :app` call in the application layout.

This is being tested with both Rails 7.2 (without Propshaft) and Rails 8 (with Propshaft) and it looks good.

Fixes #470

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,9 +101,11 @@ Then, run the `tailwindcss:upgrade` task. Among other things, this will try to r
 Here's what the upgrade task does:
 
 - Cleans up some things in the generated `config/tailwind.config.js`.
-- Runs the upstream upgrader (note: requires `npx` to run the one-time upgrade, but highly recommended).
-- Removes references to the Inter font from the application layout.
 - If present, moves `config/postcss.config.js` to the root directory.
+- If present, moves `app/assets/stylesheets/application.tailwind.css` to `app/assets/tailwind`.
+- Removes unnecessary `stylesheet_link_tag "tailwindcss"` tags from the application layout.
+- Removes references to the Inter font from the application layout.
+- Runs the upstream upgrader (note: requires `npx` to run the one-time upgrade, but highly recommended).
 
 Here's what that upgrade looks like on a vanilla Rails app:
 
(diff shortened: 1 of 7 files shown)


flavorjones:
I'm not positive this is the best possible implementation of a fix, but I'm feeling a bit of time pressure and this is good enough to ship in a second release candidate.

PR #476, opened 2025-01-26
Update README for postcss configuration for v4

EricGusmao:
For reference: 
- https://tailwindcss.com/docs/installation/using-postcss
- https://tailwindcss.com/docs/upgrade-guide#using-postcss

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,13 +251,10 @@ For example, to enable nesting:
 
 ```js
 // postcss.config.js
-module.exports = {
+export default {
   plugins: {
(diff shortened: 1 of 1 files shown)


flavorjones:
@EricGusmao Thank you.

What do you think this gem needs to do during the upgrade process with respect to this file? Currently it moves it but does not modify it. Should we attempt to modify it? Should we display a notice telling people to modify it themselves? Does the upgrade tool modify it?

flavorjones:
Amazing, thank you.

PR #478, opened 2025-01-27
Remove font assets

EricGusmao:
Since inter font is not part of this gem anymore, this PR remove the assets related to the font. Making the compiled assets cleaner, because they are included by default if you run ``rails assets:precompile`` in a v4.0.0-rc2 installation of this gem.

Captura de tela 2025-01-27 080400


flavorjones:
Thank you! I totally forgot to do this in 144eaa75

PR #480, opened 2025-01-27
doc: update README with TailwindCSS v4 upgrade instructions

EricGusmao:


README.md
```diff
@@ -91,12 +91,49 @@ gem "tailwindcss-ruby", "~> 3.4"
...
+
+```js
+  content: [
+    '../public/*.html',
+    '../app/helpers/**/*.rb',
+    '../app/javascript/**/*.js',
+    '../app/views/**/*.{erb,haml,html,slim}'
+  ],
+```
+(Just add an additional ``.`` to all the paths referenced)
```
flavorjones, line 108: I mentioned this in the Discussions thread, but I think this is a bug that should be reported upstream. If you have a clean repro, would you please report that the path is being interpreted as relative to the config file in the upgrade tool (and not CWD as the CLI correctly does)?

README.md
```diff
@@ -91,12 +91,49 @@ gem "tailwindcss-ruby", "~> 3.4"
 
 First, update to `tailwindcss-rails` v4.0.0 or higher. This will also ensure you're transitively depending on `tailwindcss-ruby` v4.
 
-``` html
+```html
 # Gemfile
 gem "tailwindcss-rails", "~> 4.0" # which transitively pins tailwindcss-ruby to v4
 ```
+To replace the class names accordingly for v4, go to ``config/tailwind.config.js`` update the ``content`` part to:
```
flavorjones, line 98: I would like to call this section out as a separate section in the doc. Some people may not care about upgrading their class names, especially if it's difficult (which this process surely will be for a large number of people).

flavorjones (changes requested):

README.md
flavorjones: Also if (when?) we automate this in the `tailwindcss:upgrade` task we can more easily remove the section later.

EricGusmao: addressed at ffe7178

flavorjones (approved):
I rebased and restructured the headers slightly. Thank you so much!

PR #481, opened 2025-01-27
Rename application.tailwind.css to application.css

flavorjones:
Its location is app/assets/tailwind so we don't need to say "tailwind" again.

Also update tests, simplify some code, update the README, consolidate the CHANGELOG.

cc @brunoprietog


pdobb:
> Its location is app/assets/tailwind so we don't need to say "tailwind" again.

While *I* can't agree with this more... my editor can. Now I'm not sure how to activate TailwindCSS syntax highlighting for app/assets/tailwind/application.css other than to go manually select "TailwindCSS" from the list each time I open the file. Or, by associating all `.css` files to the TailwindCSS syntax. Not sure if I'm the only one? I'm using Sublime Text if it matters.

PR #482, opened 2025-01-28
Fix: update postcss config path to match new structure

EricGusmao:


flavorjones:
Ugh, nice catch, thank you.

PR #483, opened 2025-01-29
Hide tailwindcss:authentication from rails g

patriciomacadden:
If you run `rails g` you can see the following output under Tailwindcss:

```
Tailwindcss:
  tailwindcss:authentication
```

You would normally run just `rails g authentication` and it will pick your template engine of choice (tailwindcss comes up as a template engine) and run that generator for you.

This PR hides it from the help text, similar to https://github.com/rails/rails/pull/53798 (does the same to `erb:authentication`).


flavorjones:
Nice! TIL about hidden generator namespaces.

EricGusmao:
By the way, this PR can be backported to v3

PR #484, opened 2025-01-29
Implement Engines support

bopm:
This PR implements support for Rails Engines Tailwind styles. It does not require configuration, but sets basic conventions and allows users to redefine engine styles in the host application.

### DIFF
--- a/lib/tailwindcss/commands.rb
+++ b/lib/tailwindcss/commands.rb
@@ -31,6 +31,40 @@ module Tailwindcss
       def rails_css_compressor?
         defined?(Rails) && Rails&.application&.config&.assets&.css_compressor.present?
       end
+
+      def engines_tailwindcss_roots
+        return [] unless defined?(Rails)
+
+        Rails::Engine.subclasses.select do |engine|
+          begin
+            spec = Gem::Specification.find_by_name(engine.engine_name)
+            spec.dependencies.any? { |d| d.name == 'tailwindcss-rails' }
+          rescue Gem::MissingSpecError
+            false
+          end
+        end.map do |engine|
+          [
+            Rails.root.join("app/assets/tailwind/#{engine.engine_name}/application.css"),
+            engine.root.join("app/assets/tailwind/#{engine.engine_name}/application.css")
+          ].select(&:exist?).compact.first.to_s
+        end.compact
+      end
+
+      def enhance_command(command)
+        engine_roots = Tailwindcss::Commands.engines_tailwindcss_roots
+        if engine_roots.any?
+          Tempfile.create('tailwind.css') do |file|
+            file.write(engine_roots.map { |root| "@import \"#{root}\";" }.join("\n"))
+            file.write("\n@import \"#{Rails.root.join('app/assets/tailwind/application.css')}\";\n")
+            file.rewind
+            transformed_command = command.dup
+            transformed_command[2] = file.path
+            yield transformed_command if block_given?
+          end
+        else
+          yield command if block_given?
+        end
+      end
     end
   end
 end
--- a/README.md
+++ b/README.md
@@ -346,6 +346,12 @@ Then you can use yarn or npm to install the dependencies.
 
 If you need to use a custom input or output file, you can run `bundle exec tailwindcss` to access the platform-specific executable, and give it your own build options.
 
+## Rails Engines support
+
+If you have Rails Engines in your application that use Tailwind CSS, they will be automatically included in the Tailwind build as long as they conform to next conventions:
+
+- The engine must have `tailwindcss-rails` as gem dependency.
+- The engine must have a `app/assets/tailwind/<engine_name>/application.css` file or your application must have overridden file in the same location of your application root.
 
 ## Troubleshooting
 
--- a/lib/tasks/build.rake
+++ b/lib/tasks/build.rake
@@ -3,8 +3,10 @@ namespace :tailwindcss do
   task build: :environment do |_, args|
     debug = args.extras.include?("debug")
     command = Tailwindcss::Commands.compile_command(debug: debug)
-    puts command.inspect if args.extras.include?("verbose")
-    system(*command, exception: true)
+    Tailwindcss::Commands.enhance_command(command) do |transformed_command|
+      puts transformed_command.inspect if args.extras.include?("verbose")
+      system(*transformed_command, exception: true)
+    end
   end
 
   desc "Watch and build your Tailwind CSS on file changes"
@@ -13,8 +15,10 @@ namespace :tailwindcss do
     poll = args.extras.include?("poll")
     always = args.extras.include?("always")
     command = Tailwindcss::Commands.watch_command(always: always, debug: debug, poll: poll)
-    puts command.inspect if args.extras.include?("verbose")
-    system(*command)
+    Tailwindcss::Commands.enhance_command(command) do |transformed_command|
+      puts transformed_command.inspect if args.extras.include?("verbose")
+      system(*transformed_command)
+    end
   rescue Interrupt
     puts "Received interrupt, exiting tailwindcss:watch" if args.extras.include?("verbose")
   end


flavorjones:
Thanks for thinking about this!

I'd like to land v4 support before I think about what's next, so it may be a day or two before I can review and give good feedback.

bopm:
Totally, this needs v4 to work. Ping me when you get to this, I am happy to add missing parts including tests.

flavorjones:
@bopm I appreciate your patience. Reviewing this PR is near, but not at, the top of my list and this is going to be a very busy week for me.

bopm:
Sure, once again, take your time. I only keeping hand on a pulse to be sure it's not blocked by you possibly waiting on something from me.

flavorjones:
@bopm Thanks for your patience, and thanks for putting together this PR!

I think this is the right general direction! But because I think our mental models aren't in sync yet, I wanted to talk through some of the things that came to mind while I read the code.

**Clarifying the use case**

In the discussion thread, there were a few use cases brought up, including (but not limited to):

1. generating a separate `tailwind.css` file for each engine ([comment](https://github.com/rails/tailwindcss-rails/discussions/355#discussioncomment-9266128), [comment](https://github.com/rails/tailwindcss-rails/discussions/355#discussioncomment-9706529))
2. generating one single `tailwind.css` that encompasses the app and all its mounted/referenced engines ([comment](https://github.com/rails/tailwindcss-rails/discussions/355#discussioncomment-9255815))

It feels like this PR is aimed at the second problem description, which is fine and is where [DHH aimed the discussion](https://github.com/rails/tailwindcss-rails/discussions/355#discussioncomment-9276200). I just want to make sure we're being clear that we're solving **that** problem and not the first one.

(And, to be even more clear, I think the first problem (gems like ActiveAdmin) can probably be solved very simply using `tailwindcss-ruby` directly. If someone has tried this and it hasn't worked properly, please drop a note in [the discussion](https://github.com/rails/tailwindcss-rails/discussions/355).)

**Finding engines**

We should probably only be looking at mounted engines, so `Rails::Engine.subclasses` is a bit too broad of a search in my opinion.

But stepping back a moment, I had originally suggested a railtie configuration for engines to explicitly register with this gem, rather than this gem having to go out and find relevant engines. What do you think of using initializers for this instead of class and file existence? 

Or, 🤯 if we _really_ want to embrace the tailwind model of "zero config", why not just add every mounted engine's root path to the tailwind search path? If the engine doesn't have tailwind CSS styles, no big whoop, tailwind seems fast enough and smart enough to figure it out, and that implementation would be simpler both for this gem but also for the engine maintainers.


**Is `@import` the right tool?**

I haven't had to include external engines yet, and so I wanted to ask: Does `@import` do what's needed here?

I had imagined that we would need to add the the engine root directory to the list of directories searched by tailwind, perhaps using the [`@source` rule](https://tailwindcss.com/docs/detecting-classes-in-source-files#explicitly-registering-sources).

Is the expectation that the application.css that we import does this source declaration itself? Seeing an end-to-end example would be really helpful to understand both the problem you're trying to solve and how this proposal would solve it, I just haven't played with it enough yet.

I realize creating an end-to-end example (preferably in an integration test suite script like we do for [installation](https://github.com/rails/tailwindcss-rails/blob/main/test/integration/user_install_test.sh) and [upgrade](https://github.com/rails/tailwindcss-rails/blob/main/test/integration/user_upgrade_test.sh)) is a lot of work, and you were probably waiting for feedback before putting in the additional effort, but I do think we're on the right path here so maybe it would be worth the investment.


**Temporary file versus editing the file that's in source control**

This PR creates a temporary file with the additional engines' configuration, and rewrites the command that is run.

I think I would prefer to modify files that are expected to be checked into source control so that it's clear what's happening and so it's easily debuggable by the app owner or engine maintainer.

The simplest pattern I can think of is to use a new file for the engine config. Add something like `@import "engines.css"` to application.css once, and then rewrite the engine config into `app/assets/tailwind/engines.css` every time the build command is run. WDYT?

---

Thanks again for this PR, I do appreciate your time and patience.

bopm:
While I'll need some time to address the whole set of questions, allow me to answer on

> Is @import the right tool?

[Here is](https://github.com/tailwindlabs/tailwindcss/pull/14373#issuecomment-2618976373) pointer towards it from a Tailwind team member. And I tend to agree. If we use `@source` instead of `@import`, we will limit engine authors' ability to decide what belongs and does not belong in the Tailwind content. 

By using `@import` we allow the engine author to use multiple `@source` directives in his `app/assets/tailwind/ENGINE_NAME/application.css` to define any needed config, including dependency on external non-rails assets for example.

flavorjones:
Thanks for helping me understand the import-vs-source distinction.

> we will limit engine authors' ability to decide what belongs and does not belong in the Tailwind content

For engine authors who don't care, though, would `@source` be appropriate? I'm asking these questions to help inform DX design tradeoffs. Specifically, this gem could either `@import` a CSS file _or_ `@source` the engine root directory based on config params or file existence (or something else).

bopm:
> we're solving **that** problem and not the first one.

Yep, we trying to generate single file from all engines and host app involved.

 > first problem (gems like ActiveAdmin) can probably be solved very simply using `tailwindcss-ruby` directly. 
 
 What I have in mind as one of examples is possible scenarios for engines extending each other. Let's say I have free engine, and commercial extension to that. I like them both processed in a orderly manner and give me styles. 


>We should probably only be looking at mounted engines, so `Rails::Engine.subclasses` is a bit too broad of a search in my opinion.

I tend to disagree here. [Here](https://github.com/full-stack-biz/turbo_material) is my another example. I am not mounting anything, but provide component library. Component libs are kinda really popular these days. Using [fact](https://github.com/rails/tailwindcss-rails/pull/484/files#diff-8ae2271570d8748367f44ba06073a1dd42bd290e48519e8ac78669a616d48920R41) that engine is dependent on our gem is good way to be specific. No reason for engine authors to include that dependency if they not produce Tailwind code.

> What do you think of using initializers for this instead of class and file existence?

Configuration instead of convention. As long as engine authors conform, host app developers don't have to do manual work for them.

> why not just add every mounted engine's root path to the tailwind search path? 

As I mentioned in previous answer, we passing the torch to engine authors. As long as they use v4 and `@source` configuration it'll be natural.

> Seeing an end-to-end example would be really helpful to understand both the problem you're trying to solve and how this proposal would solve it, I just haven't played with it enough yet.

Here is example:

- [Lib](https://github.com/full-stack-biz/turbo_material/blob/feature/tailwind-v4/app/assets/tailwind/turbo_material/application.css)
- Host application`app/assets/tailwind/application.css` head section:
```css
@import 'tailwindcss' source(none);

@source "../../javascript/**/*.js";
@source "../../../views/**/*.{erb,haml,html,slim}";

@theme {
```

>I realize creating an end-to-end example (preferably in an integration test suite script like we do for [installation](https://github.com/rails/tailwindcss-rails/blob/main/test/integration/user_install_test.sh) and [upgrade](https://github.com/rails/tailwindcss-rails/blob/main/test/integration/user_upgrade_test.sh)) is a lot of work, and you were probably waiting for feedback before putting in the additional effort, but I do think we're on the right path here so maybe it would be worth the investment.

I am going to try, as soon as we agree that overall approach is solid and on the road to be merged. 

> checked into source control

Here is problem. Bundler does not have executable callbacks on installs/updates. So you will have to run manual task on every possible bundle update to update that file. And as long as you not building everything in Docker, you may get into situation where what you committed into source control is different by gem source roots from your build environment. Ruby version change will broke it too. Too many variables in getting these engines roots. Reflection is more trusted way if you ask me. And requires no oversight.
 
>so it's easily debuggable by the app owner or engine maintainer.

Debugging is kinda easier than it may look:

```bash
bin/rails 'tailwindcss:watch[verbose]'
["/Users/sergeymoiseev/.rbenv/versions/3.4.1/lib/ruby/gems/3.4.0/gems/tailwindcss-ruby-4.0.0-arm64-darwin/exe/arm64-darwin/tailwindcss", "-i", "/var/folders/5b/2y5y0p_j575b6mpq1wvbyrz40000gn/T/tailwind.css20250219-86561-96647n", "-o", "/Users/sergeymoiseev/APP_NAME/app/assets/builds/tailwind.css", "--minify", "-w"]
```
and then in separate terminal while command is running
```bash
cat /var/folders/5b/2y5y0p_j575b6mpq1wvbyrz40000gn/T/tailwind.css20250219-86561-96647n
@import "/Users/sergeymoiseev/.rbenv/versions/3.4.1/lib/ruby/gems/3.4.0/bundler/gems/turbo_material-9f6aa17c1a44/app/assets/tailwind/turbo_material/application.css";
@import "/Users/sergeymoiseev/APP_NAME/app/assets/tailwind/application.css";
```
but we clearly can add file contents to verbose output for better DX

>The simplest pattern I can think of is to use a new file for the engine config. Add something like `@import "engines.css"` to application.css once, and then rewrite the engine config into `app/assets/tailwind/engines.css` every time the build command is run. WDYT?

Clearly we can go this way, to remove temp file processing and give developer this file in exact location to investigate even if you killed watch command or your build ended, but probably we need to agree that we adding this file to `.gitignore` by default (due to the fact that build will generate it differently on every engine update)

> @source the engine root

First thing that comes to mind, is that typical engine has dummy test app, which may have it's own content we do not want to process. Plus what do we gain even if author don't care? They still probably have sources setup in engine, and we only have to include from host.

bopm:
and @flavorjones I appreciate your feedback and need for understanding. I really hope to arrive to the level when we going to  have educated mutually agreed approach.

flavorjones:
I appreciate your thorough response!

> >What do you think of using initializers for this instead of class and file existence?
>
> Configuration instead of convention. As long as engine authors conform, host app developers don't have to do manual work for them.

Sorry, to be clear I meant the _engine_ developer would write the initializer in the engine's railtie, not the host app developer. But I get your point about conventions.

> [Here](https://github.com/full-stack-biz/turbo_material) is my another example. I am not mounting anything, but provide component library. Component libs are kinda really popular these days.

OK, seeing a live example of a non-mounted engine is really valuable, thank you for pointing me at it. I'd like the opportunity to play with this a bit to update my mental models, because the lessons you've learned are already encoded here. Let me read the code and come back to this thread better prepared.

bopm:
Quick check, @flavorjones how are your exploration is going? Can I help you with something?

flavorjones:
@bopm thanks again for your patience, this week I've not had much time to work on open source, but planning on getting back to this over the weekend.

flavorjones:
I spent some time today looking at `turbo_material` and thinking about this today. I understand and agree with many/most of your points above in [this comment](https://github.com/rails/tailwindcss-rails/pull/484#issuecomment-2669676647), and I think we're aligned on the general direction.

Give me a day to write up more complete thoughts, please. I need to ruminate on the one or two points where we're thinking differently.

Thank you for pushing this conversation forward in a healthy way!

bopm:
Thanks for clarifying, I am going to explore tests for our case and non-dynamic input file generation in a meanwhile.

bopm:
@flavorjones I've added basic tests for my functionality. I like to ask you to consider the possibility of merging if the things you are thinking about can be produced as enhancements down the line in separate PRs.

flavorjones:
@bopm I'll take a look as soon as I have time, the last few weeks have been extraordinarily busy.

bopm:
Thanks for keeping me in the loop, the idea of splitting it into phases growing from the fact that you are busy and introduced changes, in general, is not breaking and may be used as a stepping stone for other developers to start exploring their specific needs and scenarios to extend it.

bopm:
@flavorjones I apologize for bugging you, but this PR has been pending for two months. And it kinda feels that we are not getting closer to getting it in. I have the same question as before: what can I do to make it happen?

flavorjones:
@bopm Thanks for your patience. I've had very little time for open source in the last few weeks.

flavorjones:
My first request to you is: please do not merge master into PRs. Prefer rebasing, please. It makes it easier to review.

lib/tailwindcss/commands.rb
```diff
@@ -32,6 +32,40 @@ def watch_command(always: false, poll: false, **kwargs)
       def rails_css_compressor?
         defined?(Rails) && Rails&.application&.config&.assets&.css_compressor.present?
       end
+
+      def engines_tailwindcss_roots
+        return [] unless defined?(Rails)
+
+        Rails::Engine.subclasses.select do |engine|
+          begin
+            spec = Gem::Specification.find_by_name(engine.engine_name)
```
flavorjones, line 41: Is `Gem::Specification.find_by_name` guaranteed to return spec for the version of the gem that is loaded? If there are multiple versions installed on the system, could it return the spec from an earlier or later version?

I've not seen this pattern of using an engine's dependencies to control behavior, and it feels very indirect/implicit. I'm very hesitant to use this approach.

Help me understand why we shouldn't use a railtie and ask engines to be explicit/imperative? The engine, in its railtie file, should explicitly do something like:

```ruby
ActiveSupport.on_load(:tailwindcss_rails) do
  config.tailwindcss_rails.engine_roots << "path/to/application.css"
end
```

That would greatly simplify what we do in the gem.

lib/tailwindcss/commands.rb
```diff
@@ -32,6 +32,40 @@ def watch_command(always: false, poll: false, **kwargs)
...
+          end
+        end.map do |engine|
+          [
+            Rails.root.join("app/assets/tailwind/#{engine.engine_name}/application.css"),
+            engine.root.join("app/assets/tailwind/#{engine.engine_name}/application.css")
+          ].select(&:exist?).compact.first.to_s
+        end.compact
+      end
+
+      def enhance_command(command)
```
flavorjones, line 54: I don't like this pattern of modifying the command string. I would prefer if `#compile_command` and `#watch_command` accepted an input file path that defaulted to a method that creates this file and returns the string. Something like:

```ruby
      def compile_command(input: application_css, debug: false, **kwargs)
        debug = ENV["TAILWINDCSS_DEBUG"].present? if ENV.key?("TAILWINDCSS_DEBUG")
        rails_root = defined?(Rails) ? Rails.root : Pathname.new(Dir.pwd)

        command = [
          Tailwindcss::Ruby.executable(**kwargs),
          "-i", application_css,
          "-o", rails_root.join("app/assets/builds/tailwind.css").to_s,
        ]

        command << "--minify" unless (debug || rails_css_compressor?)

        postcss_path = rails_root.join("postcss.config.js")
        command += ["--postcss", postcss_path.to_s] if File.exist?(postcss_path)

        command
      end

      def application_css
        rails_root = defined?(Rails) ? Rails.root : Pathname.new(Dir.pwd)

        file = Tempfile.new("tailwind.application.css")
        engine_roots.each do |root|
          file.puts "@import \"#{root}\";"
        end
        file.puts "\n@import \"#{rails_root.join('app/assets/tailwind/application.css')}\";"
        file.close

        file.path
      end

```

flavorjones (changes requested):

bopm:
> I have a request for you: please do not merge main into a PR branch. Prefer rebasing, please. It makes it easier to review.

Rebased. Looking over standing comments.

lib/tailwindcss/commands.rb
bopm: > tempfile generation doesn't bleed into the rake tasks.
The main reason I've done it the way I did it is because of Tempfile removal. Using `Tempfile.create` block [guarantees](https://ruby-doc.org/stdlib-2.5.3/libdoc/tempfile/rdoc/Tempfile.html#method-c-create) that the file will be removed.

lib/tailwindcss/commands.rb
bopm: > Gem::Specification.find_by_name guaranteed to return spec for the version of the gem that is loaded

Yes, because it's overridden by Bundler, example:

```ruby
rails-empty(dev)> Gem::Specification.find_by_name('rails')
=>
Gem::Specification.new do |s|
  s.name = "rails"
  s.version = Gem::Version.new("8.0.2")
  s.installed_by_version = Gem::Version.new("3.6.1")
  s.authors = ["David Heinemeier Hansson"]
```

bopm:
@flavorjones I understand your concerns, but it'll take me some time to address them.

bopm:
@flavorjones I tried to address your requirements, and here is [the intermediate result](https://github.com/bopm/tailwindcss-rails/pull/1/files). It took me two evenings to implement. And I am deeply unhappy with the end result:
- complexity grown.
- I am dealing with Engine config, which was never my intention in the first place, and I am not sure what I am doing.
- Due to Engine config, tests are grown out of proportion too.
- Whole Tempfile management feels wrong.

I thinking of walking away from this problem, cutting the original approach as a monkey-patching gem on top of your gem and calling it a loss. It does not feel that I have here any support from the core team or any interest from anyone but me to get it done.

lib/tailwindcss/commands.rb
bopm: How about [this approach](https://github.com/bopm/tailwindcss-rails/pull/2/files) for solving this problem?

flavorjones:
@bopm I'm sorry you're frustrated. However, given the amount of time I've spent talking about engine support with you and other people across several threads, it's hard for me to hear you say "it does not feel like I have  ... any interest from anyone but me to get it done." Maybe you should adjust your expectations, I'm maintaining this gem mostly in my free time and doing my best to make sure any solution matches existing Rails design patterns (which I know well) and existing engine tailwind patterns (which you know better than me).

At the same time, if you feel like you want to release a separate gem, you're free to do so.

I've reviewed your separate branch and -- despite your frustration -- this is closer to what I want to see in the gem. I understand that you think it's more complex, however this is how Rails engine integration is meant to work. If you have no objection, I'd like to try to finish this up, giving you co-author credit on the commits that are taken from your suggested approach to the tempfile in this PR.

bopm:
@flavorjones I am trying to be completely transparent here. That is why I originally included in my comment both the idea of making a separate gem and my overall frustration with the process. The key moment there is not about an absence of effort. But the absence of the result. The discussion is currently 1 year old. My first attendance in it is 7 months old. This PR is 3 months old. And yet it started to feel we were no closer to "get it done" than we were 5-6 months ago.

Yet, after posting it, I let my frustration to cool down for a few days, and then get back to work with [this approach](https://github.com/bopm/tailwindcss-rails/pull/2/files). While I am feeling a bit out of my depth with Engine config, I done my fair share of dealing with temp files in Rails, and this is the hill I am ready to stand on. [This](https://github.com/bopm/tailwindcss-rails/blob/feature/another-approach/lib/tasks/build.rake#L7-L13) is safer and better than [this](https://github.com/bopm/tailwindcss-rails/blob/feature/engines-pt.2/lib/tasks/build.rake#L7-L12). If we can find middle ground in that, I will only need to fix tests for [this branch](https://github.com/bopm/tailwindcss-rails/pull/2/files) and force push it here to keep PR for history.

PR #490, opened 2025-02-04
show focus outlines

### LINKED ISSUE #489: field outlines are hidden - even when focus is applied.
A typical form field looks like the following:

```erb
    <%= form.textarea :address, rows: 4, class: ["block shadow-sm rounded-md border outline-hidden px-3 py-2 mt-2 w-full",
                    {"border-gray-400 focus:outline-blue-600": order.errors[:address].none?,
                    "border-red-400 focus:outline-red-600": order.errors[:address].any?}] %>
```

Key items:
  * `outline-hidden`
  * `focus:outline-blue-600`
  * `focus:outline-red-600`

Missing is a focus [outline-style](https://tailwindcss.com/docs/outline-style) something like solid, dashed, dotted, or double.  Not knowing the intent here, I'm hesitant to provide a PR.

rubys:
Fixes #489

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # `tailwindcss-rails` Changelog
 
+## next / unreleased
+
+Style changes to templates:
+
+* Field outlines are no longer hidden, and the focus border is brighter. #489 @rubys
+
+
 ## v4.0.0 / 2025-02-01
 
 ### Upgrade to Tailwind CSS v4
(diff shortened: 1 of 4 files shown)


patriciomacadden:
Maybe worth doing the same to the new session view?

otherwise they look different:

passwords:
image

new session:
image

the new session looks brighter because it doesn't have the `focus:outline-solid` class.

What I'm not sure is what looks better, if no `focus:outline-solid` and no `outline-hidden` at all or just `focus:outline-solid`.

what do you think?

rubys:
Other than "they probably should be consistent", I have no opinion :-)

I'm just trying to update [Agile Web Development with Rails 8](https://pragprog.com/titles/rails8/agile-web-development-with-rails-8/), and would like to get an idea of what this will ultimately look like.

PR #461, opened 2025-01-21
index: show the actions to the right of the partial and add a division

patriciomacadden:
between objects.

Currently, when your collection has many objects, and given the "show this xxx" button is grey, it's hard to distinguish when an object ends and another one begins. 

This PR adds a subtle division between objects and wraps the partial in a flex item, moving the show button to the right hand side. Also, as suggested in #457 , I've added the edit and destroy buttons.

| before | after |
| --- | --- |
| image | image |

PR #498, opened 2025-02-12
Add confirm prompt if turbo is available

patriciomacadden:
In the past, the scaffold generator added a confirm prompt to the destroy button. This PR adds it back only if Turbo is available.


patriciomacadden:
also adding `cursor-pointer` class to the destroy button as otherwise it'll show the normal cursor, which is odd for buttons.

flavorjones (commented):
@patriciomacadden Do we need to check if Turbo is available? What's the down side of always adding the `data-turbo-confirm` attribute?

It feels like unnecessary complexity, since I'm assuming that the attribute has no effect if Turbo is not present, but I'm not very familiar with turbo-rails so maybe I'm missing something.

Also, would you mind updating this PR to also include the destroy link merged from #461?

patriciomacadden:
@flavorjones i don't think there's a downside of adding `data-turbo-confirm` if turbo isn't present. It would be only a generated attribute that won't be used by anyone, but i don't think it's a problem. I thought it was clearer this way, but I can remove the check. No problem.

I'll also add the styles merged in #461 .

flavorjones:
woot!

PR #503, opened 2025-02-20
Improve views for mobile devices

patriciomacadden:
As always, here's the before/after:

| before | after |
| --- | --- |
| sessions_new | sessions_new |
| passwords_new | passwords_new |
| passwords_edit | passwords_edit |
| index | index |
| new | new |
| edit | edit |
| show | show |


flavorjones:
Oh, this is so good. Thank you!

patriciomacadden:
my idea is to keep contributing these kind of things so the scaffold generator is usable enough without many modifications as possible. will try to find more stuff to contribute :) thanks for merging!

Issue #400, opened 2024-09-11
Allow the puma plugin to enable "debug" (no minification)


richardkmichael:
It would be helpful if the puma plugin could permit configuration of the TailwindCSS watcher in debug mode (to generate non-minified output).  It seems this is only possible when starting the watcher using rake task (and therefore also a `Procfile`).

The Puma configuration DSL [`plugin` does not accept arguments other than the plug-in name](https://github.com/puma/puma/blob/master/lib/puma/dsl.rb#L168-L175), so this can't be implemented as: `plugin :tailwindcss, :debug`.

It could be done with an environment variable .  (The plug-in already uses environment for `TAILWINDCSS_INSTALL_DIR`.)

A simple patch:

```ruby
# lib/puma/plugin/tailwindcss.rb

def start(launcher)
    debug = ["true", "1"].include? ENV["TAILWINDCSS_DEBUG"]
    # ...
        IO.popen(Tailwindcss::Commands.watch_command(debug: debug), 'r+') do |io|
    # ...
end
```

flavorjones:
@richardkmichael Thanks for suggesting this addition. I'd prefer if we addded the environment variable logic to `lib/tailwindcss/commands.rb`, in the `compile_command` method.

Are you interested in putting together a PR for that? Ideally I'd also like a test added to `test/lib/tailwindcss/commands_test.rb` so we don't accidentally break it in the future.

flavorjones:
Closing, but still happy to review a pull request related to this.

PR #504, opened 2025-02-21
Tailwindcss debug environment variable

r-sierra:
Hi, this PR adds an environment variable to generate non-minified assets as discussed in #400

### DIFF
--- a/lib/tailwindcss/commands.rb
+++ b/lib/tailwindcss/commands.rb
@@ -4,7 +4,7 @@ module Tailwindcss
   module Commands
     class << self
       def compile_command(debug: false, **kwargs)
-        debug ||= ["true", "1"].include?(ENV["TAILWINDCSS_DEBUG"])
+        debug = ENV["TAILWINDCSS_DEBUG"].present? unless ENV["TAILWINDCSS_DEBUG"].nil?
         rails_root = defined?(Rails) ? Rails.root : Pathname.new(Dir.pwd)
 
         command = [
--- a/lib/tailwindcss/commands.rb
+++ b/lib/tailwindcss/commands.rb
@@ -4,7 +4,7 @@ module Tailwindcss
   module Commands
     class << self
       def compile_command(debug: false, **kwargs)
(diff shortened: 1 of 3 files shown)


test/lib/tailwindcss/commands_test.rb
```diff
@@ -33,6 +33,34 @@ def setup
     end
   end
 
+  test ".compile_command debug environment variable" do
+    begin
+      Rails.stub(:root, File) do # Rails.root won't work in this test suite
```
flavorjones, line 38: If `debug: true` is passed to `compile_command` but `TAILWINDCSS_DEBUG` is unset, which should take precedence? I *think* the environment variable probably should, but that's a weakly held opinion.

Regardless of which way you want it, I'd like test coverage for that case.

lib/tailwindcss/commands.rb
```diff
@@ -4,6 +4,7 @@ module Tailwindcss
   module Commands
     class << self
       def compile_command(debug: false, **kwargs)
+        debug ||= ["true", "1"].include?(ENV["TAILWINDCSS_DEBUG"])
```
flavorjones, line 7: Please change this to test for `ENV["TAILWINDCSS_DEBUG"].present?`

flavorjones (changes requested):
Thank you for drafting this! I'd like to simplify the environment variable check, and I'm not sure what to do if both the rake task argument is passed and the environment variable is set.

test/lib/tailwindcss/commands_test.rb
r-sierra: Since the argument `debug` is `false` by default, I cannot test if the value was explicitly set to `false` without changing the methods signature. I would rather prefer having the environment variable take precedence, as we can check if it was explicitly set in the `ENV` hash

r-sierra:
Hi @flavorjones, thanks for taking the time to review this changes. This is ready for another review.

flavorjones:
I rebased and made a small change from `unless ENV[].nil?` to `if ENV.key?()`. This looks great! Will merge when it goes green.

Thank you!

flavorjones (approved):

Issue #502, opened 2025-02-20
Use plugins with tailwind v4


bogdan:
I didn't find any way of adding a plugin in v4 setup. 
Adding `package.json` doesn't help:

``` json
{
  "name": "realphotos-marketplace",
  "packageManager": "yarn@4.5.2",
  "dependencies": {
    "daisyui": "5.0.0-beta.8"
  }
}
```

``` css
@import "tailwindcss";
@plugin "daisyui";
```

```
=> Booting Puma
=> Rails 8.0.1 application starting in development
=> Run `bin/rails server --help` for more startup options
Puma starting in single mode...
* Puma version: 6.6.0 ("Return to Forever")
* Ruby version: ruby 3.3.4 (2024-07-09 revision be1089c8ec) [arm64-darwin23]
*  Min threads: 3
*  Max threads: 3
*  Environment: development
*          PID: 25858
* Listening on http://127.0.0.1:3000
* Listening on http://[::1]:3000
Use Ctrl-C to stop
Error: Can't resolve 'daisyui' in '/Users/bogdan/makabu/brad/realphotos-marketplace/app/assets/tailwind'
Detected tailwind has gone away, stopping Puma...
- Gracefully stopping, waiting for requests to finish
=== puma shutdown: 2025-02-20 11:30:39 +0100 ===
```

Please suggest the doc on how to use plugins.

patriciomacadden:
Hi @bogdan ,

Can you share the steps to reproduce? because it's working for me.

My steps:

1. run `npm i -D daisyui@beta` (from daisyui docs)
2. add `@plugin "daisyui";` to `app/assets/tailwind/application.css`

I can see this in the console when tailwind builds the assets now:

```
/*! 🌼 daisyUI 5.0.0-beta.8 */
```

and I can see daisyui in the browser just fine.

Do you have the `package.json` and most importantly the `node_modules` folder at the root of your project?

It would be great If you can share more information.

I don't think it's this project's responsibility to inform its users how to use tailwind plugins though. Specially because there's no extra steps to do it. Maybe I'm wrong and if I am, please correct me.

bogdan:
It looks like in my case the issue is specifically with yarn and its `pnp` setup which is a new default. It doesn't use `node_modules`:

https://yarnpkg.com/features/pnp

I am not sure if this setup is something all js package managers will go, or it is never suppose to work everywhere.

I checked `npm` and it works fine.

If you don't plan to support that, you may close the issue.

[...]

PR #505, opened 2025-02-22
Explain plugins in readme

bogdan:
Updated readme to explain how to install plugins.

Discussion: https://github.com/rails/tailwindcss-rails/issues/502#issuecomment-2671620275

cc @flavorjones


flavorjones:
@bogdan This is great, thank you so much!

flavorjones:
I added a commit bumping it up to an H3 and moving it to after the build/rebuild commands. Merging!

sbounmy:
@bogdan thanks for the update

I had also to add
```js
// tailwind.config.js
module.exports = {
   ....
  plugins: [
    require("daisyui")
  ],
}
```
is it only me ?

PR #513, opened 2025-03-17
Fix hardcoded attribute in form

patriciomacadden:
In the _form partial, when the generated attribute is `attachments`, it generates a `file_field` with `multiple: true`. The error checking bits (to display colored borders) are hardcoded to `password` instead of `attribute.column_name`.


flavorjones:
Ah, we both missed this in af38df91. Thank you for fixing!

patriciomacadden:
yeah, my mistake! Thanks 😄

PR #520, opened 2025-03-19
add accept_confirm if turbo is available so system tests don't fail. fixes #519

### LINKED ISSUE #519: Tailwind scaffolding generates tests that will fail
Reproduction instructions:

```
rails new blog --css tailwind
cd blog
bin/rails generate scaffold Post title:string body:text
bin/rails db:migrate
bin/rails test:system
```

Results:

Image

Results if `--css tailwind` is omitted:

```
Running 4 tests in a single process (parallelization threshold is 50)
Run options: --seed 22878

# Running:

Capybara starting Puma...
* Version 6.6.0, codename: Return to Forever
* Min threads: 0, max threads: 4
* Listening on http://127.0.0.1:50646
....

Finished in 1.903383s, 2.1015 runs/s, 2.1015 assertions/s.
4 runs, 4 assertions, 0 failures, 0 errors, 0 skips
```

patriciomacadden:
Generated tests starting failing after we added the data turbo confirm attribute to delete links. 

This PR fixes that by monkey patching the test_unit/scaffold generator so after everything is generated it will wrap the `click_on` call with a `accept_confirm` call, only if turbo is defined.

will try to add test for this but wanted to confirm if the approach is something everyone would agree on.

Alternatives:
* add `source_paths << File.expand_path("templates", __dir__)` and copy the `system_test.rb.tt` from rails and add a guard there to check if turbo is available instead of gsubing the file in the generator
* remove `data-turbo-confirm` from the template.
* leave as is (but generated tests will fail 😞)


lib/generators/test_unit/scaffold/scaffold_generator.rb
```diff
@@ -0,0 +1,19 @@
+require "rails/generators/test_unit/scaffold/scaffold_generator"
+
+module TestUnit # :nodoc:
+  module Generators # :nodoc:
+    class ScaffoldGenerator < Base # :nodoc:
+      def fix_system_test
+        if turbo_defined?
+          gsub_file File.join("test/system", class_path, "#{file_name.pluralize}_test.rb"), "click_on \"Destroy this #{human_name.downcase}\", match: :first", "accept_confirm { click_on \"Destroy this #{human_name.downcase}\", match: :first }"
```
flavorjones, line 8: Can we make this a bit less brittle (in the case where the exact text of the system test changes) by using a more permissive regular expression? Maybe something like

```suggestion
          gsub_file File.join("test/system", class_path, "#{file_name.pluralize}_test.rb"), /(click_on.*Destroy this.*)$/", "accept_confirm { \\1 }"
```

flavorjones (changes requested):
I think this approach is the best option out of the ones you listed? I'd rather not have to maintain a copy of the system tests! I made one small suggestion to try to make the gsub_file less finicky (in case the upstream system tests change).

lib/generators/test_unit/scaffold/scaffold_generator.rb
flavorjones: Actually I tested this change and it works, so I added a commit doing it!

patriciomacadden: Much nicer! thanks!

patriciomacadden:
Yes, I feel the same. I didn't want to maintain a copy of the system test just for this.

And looks much nicer with your version of the gsub!

flavorjones:
I also just added a very simple integration test to make sure the gsub applies cleanly. I don't think I'm ready to add a full Rails system test to CI. 😅

PR #523, opened 2025-03-28
Simplify markup

patriciomacadden:
Removed two unneeded div wrappers. The one in index isn't needed because its parent is `flex`, the one in show can be avoided by adding `form_class` to the button_to. It generates cleaner partials an easier to change.


flavorjones:
Looks great, thank you!

PR #535, opened 2025-04-12
suppress caniuse-lite is outdated warning

wlipa:
Adds BROWSERSLIST_IGNORE_OLD_DATA to the environment when spawning from the Puma helper.  This avoids a 3 line warning message about stale browser data on every process startup.

I added this on the v3 branch only because that branch is expected to change rarely.  v4 sees frequent updates so likely no need for this configuration (at least until v5).

Fixes flavorjones/tailwindcss-ruby#76


wlipa:
Closing for now, I think there are going to be other spots to change.

PR #536, opened 2025-04-13
suppress caniuse-lite is outdated warning

wlipa:
Adds BROWSERSLIST_IGNORE_OLD_DATA to the environment when spawning from the Puma helper. This avoids a 3 line warning message about stale browser data on every process startup.

I added this on the v3 branch only because that branch is expected to change rarely. v4 sees frequent updates so likely no need for this configuration (at least until v5).

Fixes flavorjones/tailwindcss-ruby#76

take 2


flavorjones:
Thanks for opening this. I'll take a look this week.

flavorjones (approved):
Looks good, thank you. I renamed `Env` to `ENV` because it's not a class, but otherwise going to merge this and cut a release.

wlipa:
Thank you!  I think I forgot what language I was writing at the time.  😀

flavorjones:
Just released https://github.com/rails/tailwindcss-rails/releases/tag/v3.3.2

Thanks again!

Issue #559, opened 2025-07-01
The system test scaffold generator is failing when a rails project lacks system tests


rschultheis:
Using versions: `tailwindcss-rails 4.1.8` and `rails 8.0.2`.

Rails projects created with the `--skip-system-test --css=tailwind` combination of arguments get an exception when using the rails scaffold generator due to some code in this project.

To reproduce:

```
rails new --skip-system-test --css=tailwind bugdemo
cd bugdemo
rails generate scaffold demobug
```

That results in this stack trace:

```
rails generate scaffold demobug
      invoke  active_record
      create    db/migrate/20250701122835_create_demobugs.rb
      create    app/models/demobug.rb
      invoke    test_unit
      create      test/models/demobug_test.rb
      create      test/fixtures/demobugs.yml
      invoke  resource_route
       route    resources :demobugs
      invoke  scaffold_controller
      create    app/controllers/demobugs_controller.rb
      invoke    tailwindcss
      create      app/views/demobugs
      create      app/views/demobugs/index.html.erb
      create      app/views/demobugs/edit.html.erb
      create      app/views/demobugs/show.html.erb
      create      app/views/demobugs/new.html.erb
      create      app/views/demobugs/_form.html.erb
      create      app/views/demobugs/_demobug.html.erb
      invoke    resource_route
      invoke    test_unit
      create      test/controllers/demobugs_controller_test.rb
        gsub      test/system/demobugs_test.rb
/Users/rob/.rbenv/versions/3.4.4/lib/ruby/gems/3.4.0/gems/thor-1.3.2/lib/thor/actions/file_manipulation.rb:271:in 'IO.binread': No such file or directory @ rb_sysopen - /Users/rob/workspace/bugdemo/test/system/demobugs_test.rb (Errno::ENOENT)
	from /Users/rob/.rbenv/versions/3.4.4/lib/ruby/gems/3.4.0/gems/thor-1.3.2/lib/thor/actions/file_manipulation.rb:271:in 'Thor::Actions#gsub_file'
	from /Users/rob/.rbenv/versions/3.4.4/lib/ruby/gems/3.4.0/gems/tailwindcss-rails-4.2.3/lib/generators/test_unit/scaffold/scaffold_generator.rb:8:in 'TestUnit::Generators::ScaffoldGenerator#fix_system_test'
[...]

PR #560, opened 2025-07-02
check for system test existence before modifying

rschultheis:
fixes https://github.com/rails/tailwindcss-rails/issues/559

With this change the scaffold generator no longer assumes the existence of a system test file. Rails projects can be initialized without system tests using the --skip-system-test option.


rschultheis:
#561 is a better approach.  If the maintainers feel differently then please reopen.

PR #571, opened 2025-08-21
Fix Puma plugin entry point issue

### LINKED ISSUE #567: Error loading engine CSS when using Puma plugin
I'm not sure why this happens, but if you use the [Puma plugin](https://github.com/rails/tailwindcss-rails#puma-plugin), you'll get the following error.

`Error: Can't resolve '../builds/tailwind/[engine_name].css'`

If you run `rails tailwindcss:build` and then restart the rails server with the plugin, things will work. Also, if you just run `rails tailwindcss:watch`, the issue does not happen.

bopm:
Fixes #567.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Improved
 
 * Support Rails 8.1 scaffolding which disables system tests by default. #585 @flavorjones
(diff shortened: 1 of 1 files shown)


flavorjones:
Thank you @bopm!

PR #577, opened 2025-08-30
Allow customizing input/output via env variables

thewatts:
The Rails `assets:precompile` task internally calls `bin/rails tailwindcss:build`, which uses hardcoded input/output file paths. Previously, the only way to customize these paths was to bypass the Rails integration entirely and use the Tailwind CLI directly (via `bundle exec tailwindcss`).

This change introduces TAILWINDCSS_INPUT_FILE and TAILWINDCSS_OUTPUT_FILE environment variables that allow users to override the default paths while still using the standard `assets:precompile` workflow.

This is useful for projects with non-standard asset structures or when integrating with custom build pipelines.


manuelmeurer:
Hi @flavorjones, I just found this PR, since I have the same wish as @thewatts to be able to customize the input and output file(s). 😄 
My use-case is that my app has several "namespaces", which map to separate URLs and have separate styling.
Thus I'd like to have separate CSS files, one for each namespace, to allow for customizing TW and plugins, add custom CSS etc.

I understand that this is not a trivial change, since as you mentioned the default paths are assumed at different locations, but maybe it's worthwile adding this customization nevertheless?
For now, I'm using the TW CLI instead of the Rake tasks, but of course that's not optimal.

PR #578, opened 2025-09-04
Rename `on_stopped` callback to `after_stopped` for Puma 7

rono23:
- https://github.com/puma/puma/blob/master/History.md?utm_source=chatgpt.com#700
- https://github.com/puma/puma/pull/3438

image


wlipa:
In Puma 7.0.1, this is now a deprecation warning instead of a crash.

wlipa:
I'm on v4 now (partly due to this) so I won't be submitting a PR for v3.  Thanks!

Issue #477, opened 2025-01-27
v4.0.0-rc2 - application.tailwind.css is not ignored by propshaft when using :app or :all flag in stylesheet_link_tag helper


EricGusmao:
I am using rails 8.0.1, and after following the steps [here](https://github.com/rails/tailwindcss-rails/discussions/450#discussioncomment-11938025) using V4.0.0-rc2, the behaviour reported at #470  still persists. The script is running fine, the file moved directories, but it still is compiled and served by propshaft.

![Image](https://github.com/user-attachments/assets/b01b4ded-cfe3-4c28-b6ec-48c3cb6ad129)

![Image](https://github.com/user-attachments/assets/13c4e151-b4f5-412f-a120-9081d74de2b3)

This happen due to propshaft ``stylesheet_link_tag`` helper [Reference](https://github.com/rails/propshaft/blob/main/lib/propshaft/helper.rb).

I just noticed the rake task to upgrade just removes the former stylesheet entry and keeps

```ruby
<%= stylesheet_link_tag :app, "data-turbo-track": "reload" %>
```
So all the assets that ends with ``.css`` are served.

But If you also remove the line above and add (as the install script):

```ruby
<%= stylesheet_link_tag "tailwind", "data-turbo-track": "reload" %>
```

The tailwind unprocessed file is ignored and not served.

flavorjones:
🤔 This was originally fixed in 31621ead, but something I did in the meantime made this regress. Will investigate.

flavorjones:
Darn, this is the change that needs to be made, I used `after` when I should have used `before`. I'll open a PR.

```patch
diff --git a/lib/tailwindcss/engine.rb b/lib/tailwindcss/engine.rb
index 0c020bbc..7b88c5f1 100644
--- a/lib/tailwindcss/engine.rb
+++ b/lib/tailwindcss/engine.rb
@@ -6,7 +6,7 @@ class Engine < ::Rails::Engine
       Rails.application.config.generators.stylesheets = false
     end

-    initializer "tailwindcss.exclude_asset_path", after: "propshaft.append_assets_path" do
+    initializer "tailwindcss.exclude_asset_path", before: "propshaft.append_assets_path" do
       if Rails.application.config.assets.excluded_paths # the app may not be using Propshaft
         Rails.application.config.assets.excluded_paths << Rails.root.join("app/assets/tailwind")
       end
```

aurum-ellt:
[...]

Issue #565, opened 2025-07-15
Engine entry files cause issues with stylesheet_link_tag(:app)


justinfrench:
I've been playing around with the [new Engines integration](https://github.com/rails/tailwindcss-rails/pull/554), and have discovered an issue.

I've [created an issue on propshaft](https://github.com/rails/propshaft/issues/242) with more detail, but I wanted to cross-post here for awareness.

TLDR:

* A fresh Rails app uses `stylesheet_link_tag(:app)`, which adds in *all* CSS files in `app/assets/`, including files in `app/assets/builds/`, which includes the entry files generated (eg `app/assets/builds/tailwind/<engine_name>.css`).
* This means the entry files are served-up to the browser raw, rather than *after* being explicitly added to `tailwind/application.css`

I first noticed this when I saw an error in the browser console, because the file had a reference to the full local path on disk:

```
/* DO NOT MODIFY THIS FILE, it was auto-generated by tailwindcss-rails */

@import "/Users/whatever/my_gem/app/assets/tailwind/my_gem/engine.css";
```

bopm:
I tend to agree with @rafaelfranca on this one. We can try to move build artefacts out of the `builds`, but TailwindCSSes are full of magic and should not be processed by catch-all instruments like `stylesheet_link_tag :app`. Solution to create manual file and include all propshaft-related things in it, excluding Tailwind, seems more logical. But I am open to suggestions.

justinfrench:
The simplest things I can think of are:

* Update the Rails generator to avoid using `:all` with `--css=tailwind`, and/or
* Move the `:all` functionality from propshaft into rails (with documentation) so it's more discoverable as a root cause

But those are clearly not things that tailwindcss-rails an solve for.

I don't understand enough about the internals, but could we avoid all of this by **saving the entry to somewhere outside of the host app's asset path**?

justinfrench:
Yep, I’m happy to have a go at updating the docs, but if someone else gets to it first, that’s fine.

flavorjones:
@bopm 

> doing differently would make it more than a single PR here, and would push actual arrival of that feature to another half a year in the future at least

You have publicly framed my reviews on your early PRs like this several times now. If you want to discuss this with me, let's take it offline.

I don't want to avoid discussing improvements to the current (experimental) implementation for engine support. If it would be better to change the approach on engines, let's talk about it.

bopm:
Be my guest, my comment was about my personal capacity.

[...]

PR #607, opened 2026-03-04
Filter engine stubs from Propshaft :app stylesheet resolution

steveclarke:
I've been building a Rails engine that ships Tailwind CSS using the `tailwindcss:engines` mechanism from #554, and I ran into an issue with `stylesheet_link_tag :app`.

## What's happening

When the host app uses `stylesheet_link_tag :app` (the default in Rails 8.1 layouts), Propshaft globs `app/assets/**/*.css` and generates a `<link>` tag for each match. This picks up the engine stub files at `app/assets/builds/tailwind/<engine>.css` in addition to the compiled `tailwind.css`.

The browser receives the stub, which contains an `@import` with an absolute filesystem path:

```css
@import "/Users/me/.../my_engine/app/assets/tailwind/my_engine/engine.css";
```

The browser interprets that as a URL and gets a 404. The compiled `tailwind.css` already has everything inlined, so nothing is actually broken, but you get console errors on every page load.

**Reproduction**: https://github.com/steveclarke/tailwindcss-rails-engine-stub-repro

Clone it, `bundle install`, `bin/rails tailwindcss:build`, `bin/rails server`, open the console.

## Verified with the repro app

**Stock `tailwindcss-rails` 4.4.0** — three stylesheet links, stub served to browser:

```html
<link rel="stylesheet" href="/assets/application-8b441ae0.css" />
<link rel="stylesheet" href="/assets/tailwind-8272e5b1.css" />
<link rel="stylesheet" href="/assets/tailwind/mini_engine-8dad903a.css" />  <!-- the stub -->
```

**With this fix** — stub is filtered, only two links:

```html
<link rel="stylesheet" href="/assets/application-8b441ae0.css" />
<link rel="stylesheet" href="/assets/tailwind-8272e5b1.css" />
```

## Why excluded_paths doesn't work here

I initially tried adding `app/assets/builds/tailwind` to `excluded_paths`, similar to how `app/assets/tailwind` is already excluded. But `excluded_paths` only filters top-level entries from the load path array — it does exact string matching against directories like `app/assets/builds`. Since `app/assets/builds/tailwind` is a subdirectory within that load path, it doesn't match and nothing gets excluded.

## The fix

This PR adds a filter on `Propshaft::Helper#app_stylesheets_paths` that rejects assets whose logical paths start with `tailwind/`. These are always engine stubs (generated by `tailwindcss:engines` into `app/assets/builds/tailwind/`). The compiled output has the logical path `tailwind.css` (no slash), so it passes through fine.

The filter is guarded with `defined?(Propshaft::Helper)` so it's a no-op if the app uses Sprockets.

This is the same general approach as the existing `excluded_paths` initializer — just at a different layer, since Propshaft doesn't support subdirectory exclusions.

Not sure if there's a cleaner way to handle this, but this is the best I could figure out after reading through the Propshaft source. Happy to adjust if there's a better approach.

Related: #477 fixed the same class of issue for `app/assets/tailwind/` (the source directory). This covers the stubs directory that was introduced later by #554.

### DIFF
--- a/lib/tailwindcss/engine.rb
+++ b/lib/tailwindcss/engine.rb
@@ -12,6 +12,32 @@ module Tailwindcss
       end
     end
 
+    # Filter engine CSS stubs from Propshaft's :app stylesheet resolution.
+    #
+    # The tailwindcss:engines task generates intermediate stub files at
+    # app/assets/builds/tailwind/<engine>.css containing @import directives
+    # with absolute filesystem paths. These are consumed by the Tailwind CLI
+    # at build time and inlined into the compiled tailwind.css output.
+    #
+    # When hosts use stylesheet_link_tag :app (the Rails 8.1 default),
+    # Propshaft globs app/assets/**/*.css and serves these stubs directly
+    # to the browser. The browser interprets the absolute path as a URL
+    # and gets a 404.
+    #
+    # Propshaft's excluded_paths can't help here — it only filters top-level
+    # load path directories, not subdirectories within a load path.
+    initializer "tailwindcss.filter_engine_stubs" do
+      ActiveSupport.on_load(:action_view) do
+        if defined?(Propshaft::Helper)
+          Propshaft::Helper.prepend(Module.new {
+            def app_stylesheets_paths
+              super.reject { |path| path.start_with?("tailwind/") }
+            end
+          })
+        end
+      end
+    end
+
     config.app_generators do |g|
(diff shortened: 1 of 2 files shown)


flavorjones:
Please note there's discussion about this at https://github.com/rails/tailwindcss-rails/issues/565 which needs to be resolved before agreeing upon a solution. Please jump in there if you have an opinion!

PR #627, opened 2026-06-23
Add TAILWINDCSS_SKIP_BUILD to skip the Tailwind build during asset compilation

cgmoore120:
## Summary

The gem unconditionally enhances `assets:precompile` (and the test prepare task) with `tailwindcss:build`, so the Tailwind build runs automatically before assets are digested or tests are prepared. There is currently no supported way to opt out of this.

This adds a `TAILWINDCSS_SKIP_BUILD` environment variable: when set to any non-blank value, the automatic enhancement is skipped. Default behavior is unchanged when the variable is unset or blank, so this is fully backward-compatible.

## Motivation

A few situations where running the Tailwind build automatically is undesirable:

1. **Pre-built / committed CSS** — Some apps build and commit `app/assets/builds/tailwind.css` ahead of time and don't want `assets:precompile` to rebuild (and potentially overwrite) it.
2. **CI / Docker layer caching** — When the build output is produced in an earlier, cached step, re-running it during `assets:precompile` wastes time and defeats layer caching.
3. **Cross-architecture Docker builds** — When building an image for a different architecture (e.g. an `arm64` host producing an `amd64` image via emulation), the standalone CLI can fail to produce correct output. There is a known float-truncation issue in this area on Apple Silicon / Rosetta with Bun (see [oven-sh/bun#19677](https://github.com/oven-sh/bun/issues/19677) and [tailwindlabs/tailwindcss#17728](https://github.com/tailwindlabs/tailwindcss/issues/17728)). This isn't a bug in this gem, but being able to skip the automatic build lets users build the CSS in a context where it works and skip it where it doesn't.

## Precedent

The gem already reads an environment flag in the same idiom: `ENV["TAILWINDCSS_DEBUG"].present?` in `lib/tailwindcss/commands.rb`. This change uses the same `.present?` check for consistency.

## Changes

- Wrap the trailing `enhance(...)` block in `lib/tasks/build.rake` in `unless ENV["TAILWINDCSS_SKIP_BUILD"].present?`.
- Document the variable in the README near the existing build/`TAILWINDCSS_DEBUG` docs.
- Add a CHANGELOG entry.
- Add a test verifying that `tailwindcss:build` is attached to `assets:precompile` and `test:prepare` by default, that it is not attached when `TAILWINDCSS_SKIP_BUILD` is set, and that a blank value does not skip the build.

## Testing

`bundle exec rake test` passes (30 runs, 214 assertions, 0 failures, 0 errors, 0 skips).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # `tailwindcss-rails` Changelog
 
+## main (unreleased)
+
+### Improved
+
+* Setting the `TAILWINDCSS_SKIP_BUILD` environment variable to a non-blank value skips attaching `tailwindcss:build` to `assets:precompile` (and test prepare). This is useful when the CSS is pre-built/committed, when caching CI or Docker layers, or for cross-architecture Docker builds. @cgmoore120
+
+
 ## v4.6.0 / 2026-06-17
 
 ### Fixed
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@ README](https://github.com/rails/tailwindcss-rails/tree/v3-stable?tab=readme-ov-
   * [Configuration and commands](#configuration-and-commands)
   * [Building for production](#building-for-production)
   * [Building for testing](#building-for-testing)
+  * [Skipping the automatic build](#skipping-the-automatic-build)
   * [Building unminified assets](#building-unminified-assets)
   * [Live rebuild](#live-rebuild)
   * [Using Tailwind plugins](#using-tailwind-plugins)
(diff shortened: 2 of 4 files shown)


flavorjones:
@cgmoore120 Thanks for the pull request. I'm not opposed to doing something to omit tailwind compilation, but I do want to make sure we're fixing it in the right place. I am very curious why the tailwind assets are being treated differently from other assets with respect to points 1 and 2 above:

1. Why commit generated assets to source control?
2. What is the advantage to generating tailwind assets in an earlier image layer?

I don't _think_ these are common idioms with Rails, but if I'm wrong then my instinct would be to start a discussion on whether the `assets:precompile` task itself should be skippable. Do these questions make sense given that context?

